PR #2, opened 2021-09-02
Jenkins: bring back Jenkins pipeline for the tools

spcaipers-arm:
We still need to run some stages that are not available on GitHub
Actions.

E.g.:
- Coverity
- OTG-Integration Tests

Also, jenkins files will not trigger GitHub Actions

### DIFF
--- a/.github/workflows/buildmgr.yml
+++ b/.github/workflows/buildmgr.yml
@@ -14,6 +14,7 @@ on:
      - 'libs/rtemodel/**'
      - 'libs/rtefsutils/**'
      - 'tools/buildmgr/**'
+     - '!tools/buildmgr/jenkins/**'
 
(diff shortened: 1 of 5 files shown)


soumeh01 (approved):

soumeh01 (commented):
If I understand correctly, The intent here is to push 
- scripts/coverity2jenkins.py
- tools/packgen/jenkins/pipeline.groovy
- tools/buildmgr/jenkins/pipeline.groovy

Files to only **PUBLIC** open-cmsis-pack repo. Am I right?

If yes, Then I have the following questions
1. As these files will not be of any use in the public repo but will be used internally. Do we have any way to keep this stuff locally to us?
2. We should add a copyright notice to these files as well.

PR #4, opened 2021-09-07
packgen: update errors/warnings handling

brondani:
- return an error when 'packs' mandatory section is missing
- return a warning when multiple 'components' sections are defined
- print all packchk results, including errors and warnings
- avoid calling CMake when no components are defined and verbose mode is disabled

### DIFF
--- a/tools/packgen/src/PackGen.cpp
+++ b/tools/packgen/src/PackGen.cpp
@@ -92,29 +92,33 @@ int PackGen::RunPackGen(int argc, char *argv[]) {
     return 1;
   }
 
-  // Check CMake
-  result = generator.ExecCommand("cmake --version");
-  if (result.second) {
-    cerr << "packgen error: CMake was not found" << endl;
-    return 1;
-  }
+  // Avoid CMake calls when packs to be generated don't contain components and verbose mode is disabled
+  if (!generator.m_noComponents || generator.m_verbose) {
 
-  // Create CMake File API query
-  if (!generator.CreateQuery()) {
-    cerr << "packgen error: query creation failed" << endl;
-    return 1;
-  }
+    // Check CMake
+    result = generator.ExecCommand("cmake --version");
+    if (result.second) {
+      cerr << "packgen error: CMake was not found" << endl;
+      return 1;
+    }
 
-  // Parse CMake File API reply
-  if (!generator.ParseTarget()) {
-    cerr << "packgen error: target parsing failed" << endl;
-    return 1;
-  }
+    // Create CMake File API query
+    if (!generator.CreateQuery()) {
+      cerr << "packgen error: CMake File API query creation failed" << endl;
+      return 1;
+    }
 
-  // Elaborate component contents
-  if (!generator.CreateComponents()) {
-    cerr << "packgen error: component generation failed" << endl;
-    return 1;
+    // Parse CMake File API reply
+    if (!generator.ParseReply()) {
+      cerr << "packgen error: CMake File API reply parsing failed" << endl;
+      return 1;
+    }
+
+    // Elaborate component contents
+    if (!generator.CreateComponents()) {
+      cerr << "packgen error: component generation failed" << endl;
+      return 1;
+    }
   }
 
   // Create PDSC file and copy pack files
@@ -164,6 +168,10 @@ bool PackGen::ParseManifest(void) {
 
     // Pack
     YAML::Node packs = manifest["packs"];
+    if (!packs.IsDefined()) {
+      cerr << "packgen error: mandatory 'packs' key is not defined!" << endl;
+      return false;
+    }
     for (const auto& item : packs) {
       packInfo pack;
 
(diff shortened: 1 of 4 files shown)


tools/packgen/src/PackGen.cpp
```diff
@@ -277,12 +285,20 @@ void PackGen::ParseManifestApis(const YAML::Node node, packInfo& pack) {
...
+    cerr << "packgen warning: multiple 'components' sections were defined, data can be lost" << endl;
+  }
+
   YAML::Node components = node["components"];
   for (const auto& item : components) {
     const string& name = item["name"].as<string>();
 
     if (m_components.find(name) != m_components.end()) {
-      cerr << "packgen warning: component '" << name << "' has been defined multiple times" << endl;
+      cerr << "packgen warning: component '" << name << "' was defined multiple times" << endl;
```
jkrech, line 301: I would use present tense "is defined multiple times" as this is what is present in the yaml file.

tools/packgen/src/PackGen.cpp
```diff
@@ -92,29 +92,33 @@ int PackGen::RunPackGen(int argc, char *argv[]) {
...
   }
 
-  // Check CMake
-  result = generator.ExecCommand("cmake --version");
-  if (result.second) {
-    cerr << "packgen error: CMake was not found" << endl;
-    return 1;
-  }
+  // Avoid CMake calls when packs to be generated don't contain components and verbose mode is disabled
+  if (!generator.m_noComponents || generator.m_verbose) {
```
jkrech, line 96: I am unsure why CMake should be run if in verbose mode even if no components are described. Just assume the cmake command is not specified for the generation of this pack.
I seem to be missing something.

tools/packgen/src/PackGen.cpp
```diff
@@ -92,29 +92,33 @@ int PackGen::RunPackGen(int argc, char *argv[]) {
...
+      return 1;
+    }
 
-  // Elaborate component contents
-  if (!generator.CreateComponents()) {
-    cerr << "packgen error: component generation failed" << endl;
-    return 1;
+    // Parse CMake File API reply
+    if (!generator.ParseTarget()) {
+      cerr << "packgen error: target parsing failed" << endl;
```
jkrech, line 113: Do we need to think about Error Numbers and adding more detailed documentation soon?
I like your brief error message but would prefer a bit more detail e.g. the name of the CMake target or maybe even the cmake command?
e.g.
'parsing of CMake file API response for target %s failed'

tools/packgen/src/PackGen.cpp
```diff
@@ -164,6 +168,10 @@ bool PackGen::ParseManifest(void) {
 
     // Pack
     YAML::Node packs = manifest["packs"];
+    if (!packs.IsDefined()) {
+      cerr << "packgen error: mandatory 'packs' section is not defined!" << endl;
```
jkrech, line 172: I am not familiar with yaml and therefore unsure whether 'packs' is a key or 'section'?

tools/packgen/src/PackGen.cpp
```diff
@@ -277,12 +285,20 @@ void PackGen::ParseManifestApis(const YAML::Node node, packInfo& pack) {
 }
 
 bool PackGen::ParseManifestComponents(const YAML::Node node, packInfo& pack) {
+  uint32_t nodes = CountNodes(node, "components");
+  if (nodes >= 1) {
+    m_noComponents = false;
+  }
+  if (nodes > 1) {
+    cerr << "packgen warning: multiple 'components' sections were defined, data can be lost" << endl;
```
jkrech, line 293: sections or keys?

tools/packgen/test/data/CMakeTestProject/invalid_manifest2.yml
```diff
@@ -0,0 +1,65 @@
+### CMSIS Pack manifest file ###
```
jkrech, line 1: Could you add a comment to the yaml file that specifies what makes this file an invalid one?

jkrech (changes requested):

tools/packgen/src/PackGen.cpp
brondani: The verbose mode prints CMake target information and we recommend to use this feature when starting developing new packs (initially without components):
https://github.com/Open-CMSIS-Pack/devtools/tree/main/tools/packgen/docs#hints-for-generating-packs
Preventing CMake calls would break this feature.
Feel free to propose better solutions.

tools/packgen/src/PackGen.cpp
brondani: I believe we should first consolidate the errlog library or to find a replacement before doing changes in this field.
The error message above is not valid anymore, ParseTarget prints warnings with relevant information inside the function body itself. I will clean it.

tools/packgen/src/PackGen.cpp
jkrech: Understood. Let's leave it for now.

jkrech (approved):
Looks good to me.

PR #5, opened 2021-09-07
GH Action: attach cbuild installer to release

brondani:


.github/workflows/buildmgr.yml
```diff
@@ -204,6 +208,16 @@ jobs:
           retention-days: 1
           if-no-files-found: error
 
+      - name: Attach installer to release
```
spcaipers-arm, line 211: Is the `tools/buildmgr/cbuildgen/installer/cbuild_install.sh` OS dependent? If not, then I think you should do it in a single VM (using the `if` statement).

brondani: It's a step of the `create_installer` job which runs only on ubuntu-20.04

spcaipers-arm (approved):

PR #8, opened 2021-09-09
packgen: create release archive

brondani:


.github/workflows/packgen.yml
```diff
@@ -87,15 +87,54 @@ jobs:
           retention-days: 1
           if-no-files-found: error
 
-      - name: Attach packgen to release
-        if: ${{ github.event_name == 'release' }}
+  release:
```
spcaipers-arm, line 90: Would be better to add the release job in the end of the workflow (e.g. after testing?). As it is you will publish the release after the build but before the testing.

spcaipers-arm: Also, I think it should depend on tests too to pass.

jkrech (approved):
LGTM

PR #11, opened 2021-09-10
buildmgr: Fix files and components layer assignment matching recognition

brondani:


jkrech (approved):
Defect fix for layer extraction with test.

PR #23, opened 2021-09-15
Fix issues with filesystem::perms in Windows-GCC

JonatanAntoni:
On Windows a file/dir with full access permissions seems not to have
perms::all set. This may be caused by permissions available on POSIX
but not on Windows. Instead of testing file/dirs with full access for
perms::all the test checks specifically for the permissions of
interest.

Fixing usage of gtest EXPECT_EQ for better readability of error messages.

### DIFF
--- a/docs/ADDING_UNIT_TESTS.md
+++ b/docs/ADDING_UNIT_TESTS.md
@@ -91,4 +91,4 @@ are linked together. For example
 
 # Building and running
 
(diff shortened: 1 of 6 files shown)


edriouk (approved):

edriouk (approved):

jkrech (approved):
LGTM

PR #32, opened 2021-09-27
GitHub: Add CodeQL analysis

JonatanAntoni:


jkrech (dismissed):
This looks good to me. I only wonder why we choose '4' as the number of parallel builds.

Issue #35, opened 2021-09-29
Build flag order needs to be kept


slhultgren:
Currently there is no guarantee that e.g. "-I incpath2 -I incpath1" will be passed in that order when building since a set<> is used internally.

The build system first needs to be reliable, predictable and deterministic.

brondani:
Hi @slhultgren,

Thanks for the PR.
Do you have a concrete use case to describe the problem you want to tackle?

I believe we should separate the discussion in two items:

_1. The user having control over include path order_
Please note the order of components process in RTE Model is given by their IDs, the PR is not really considering it. `std::set` is also used to filter out duplicated elements, it's not rare different components providing the same include paths. Changes in this field could imply in other specification/architectural changes. In any case it is reproducible and deterministic.

_2. The generation of CMakelists.txt from cprj being deterministic_
I would agree the final sorting of fully specified include paths may not be deterministic if the underlying filesystem/pack repository change location. But to correctly address it the PR shall be amended.

Thanks
Daniel

slhultgren:
Hi @brondani 

For GCC, the last flag is the one that wins, for example if -O3 and -O0 are both specified.

For GCC, the order in which the include paths are searched is determined by the order of the include path flags on the command line.
Therefore the user needs to be able to specify the exact order of the include paths to be able to build their code.

I'll add some comments to the PR as well.

KR, Samuel

brondani:
Hi @slhultgren,

The PDSC specification already provides means to avoid header files collision by setting the `path` file attribute which can act as header namespace:
https://www.keil.com/pack/doc/CMSIS/Pack/html/pdsc_components_pg.html#element_file
Are there other use cases in which the user may need to specify the exact order of the include paths?

Cheers
Daniel

slhultgren:
Hi @brondani 

Yes, consider the case of user code in the .cprj. The user must be able to say in the cprj in which order the include paths should be used for the user code.

Regarding components I believe a change in the order of the components in the CPRJ should also be reflected in the order of include paths on the command line.
(assuming the components used specifies/exposes include paths to the user code)

KR, Samuel

slhultgren:
Hi @brondani thanks :)

Makes sense, I think I agree regarding the ordering of component "resolving". It could be tricky indeed.

[...]

PR #36, opened 2021-09-29
[rtemodel, buildmgr] Keep flag and include order

slhultgren:
Fixes issue #35

Signed-off-by: Samuel HULTGREN <samuel.hultgren@st.com>


tools/buildmgr/cbuild/src/CbuildModel.cpp
```diff
@@ -735,64 +735,41 @@ const bool CbuildModel::EvalDeviceName() {
...
-    list.insert(add_item);
+    list.push_back(add_item);
   }
   for (auto rem_item : remove) {
-    for (auto ref_item : list) {
-      if (rem_item.compare(ref_item) == 0) {
-        list.erase(ref_item);
-        break;
-      }
+    auto first_match = std::find_if(list.cbegin(), list.cend(), [rem_item](auto i){ return i == rem_item; });
```
brondani, line 770: It requires `#include <algorithm>`

brondani (changes requested):
For the time being I would avoid changes in `rtemodel` because it is already reproducible and deterministic as it is.

For completeness and consistency I would suggest to convert other `std::set` variables declared in [BuildSystemGenerator.h](https://github.com/slhultgren/devtools/blob/issue_35_build_flag_order/tools/buildmgr/cbuildgen/include/BuildSystemGenerator.h#L91-L95) and  [CbuildModel.h](https://github.com/slhultgren/devtools/blob/issue_35_build_flag_order/tools/buildmgr/cbuild/include/CbuildModel.h#L386-L388)

In addition golden references in the [test input](https://github.com/slhultgren/devtools/tree/issue_35_build_flag_order/tools/buildmgr/test/testinput/Examples) will need to be updated.

tools/buildmgr/cbuild/src/CbuildModel.cpp
slhultgren: Ok, so if I understand correctly the PR needs to also fix
* RTE Component order processing should be changed to reflect the order specified from the CPRJ?
* PR should also be amended to use list<> instead of set<> https://github.com/Open-CMSIS-Pack/devtools/blob/main/tools/buildmgr/cbuildgen/include/BuildSystemGenerator.h#L91 
* #include <algorithm>
* Update tests

brondani: Right but please leave changes in the `rtemodel` for a separate PR for better review and evaluation.

slhultgren:
@brondani  Seems there are some failing tests:

[  FAILED  ] CBuildGCCTests.TrustZone_GCC
/home/runner/work/devtools/devtools/build/testoutput/testdata/Examples/GCC/TrustZone/CM33_ns/IntDir/CMakeLists.txt is different from /home/runner/work/devtools/devtools/build/testoutput/testdata/Examples/GCC/TrustZone/CM33_ns/CMakeLists.txt.ref

[  FAILED  ] CBuildGCCTests.PreInclude
/home/runner/work/devtools/devtools/build/testoutput/testdata/Examples/GCC/Pre Include/IntDir/CMakeLists.txt is different from /home/runner/work/devtools/devtools/build/testoutput/testdata/Examples/GCC/Pre Include/CMakeLists.txt.ref

How can I see what the issue is here?

brondani:
@slhultgren In your environment after building the integration tests executable [CbuildIntegTests](https://github.com/slhultgren/devtools/tree/issue_35_build_flag_order#run-build) and setting the required [environment variables](https://github.com/slhultgren/devtools/tree/issue_35_build_flag_order#run-tests) if needed, you can selectively run only the failing tests by using the [--gtest_filter](https://google.github.io/googletest/advanced.html#running-a-subset-of-the-tests) option, for example:
```
./tools/buildmgr/test/integrationtests/windows64/Debug/CbuildIntegTests.exe --gtest_filter=CBuildGCCTests.TrustZone_GCC:CBuildGCCTests.PreInclude
```
And then compare the generated CMakeLists.txt with the expected CMakeLists.txt.ref.
I can anticipate your generated CMakeLists.txt files have duplicated include and/or pre-include entries that were being previously automatically filtered by the `std::set` containers.

tools/buildmgr/cbuildgen/src/BuildSystemGenerator.cpp
```diff
@@ -119,24 +123,24 @@ bool BuildSystemGenerator::Collect(const string& inputFile, const CbuildModel *m
...
   for (auto obj : model->GetObjects())
   {
-    m_libFilesList.insert(StrNorm(obj));
+    m_libFilesList.push_back(StrNorm(obj));
   }
 
   for (auto def : model->GetDefines())
   {
-    m_definesList.insert(def);
+    m_definesList.push_back(def);
```
brondani, line 136:
```suggestion
    CbuildUtils::PushBackUniquely(m_definesList, StrNorm(def));
```

tools/buildmgr/cbuildgen/src/BuildSystemGenerator.cpp
```diff
@@ -100,12 +100,16 @@ bool BuildSystemGenerator::Collect(const string& inputFile, const CbuildModel *m
 
   for (auto inc : model->GetIncludePaths())
   {
-    m_incPathsList.insert(StrNorm(inc));
+    std::string incpath = StrNorm(inc);
+    // Deduplicate adjacent identical include paths
+    if (m_incPathsList.empty() || m_incPathsList.back() != incpath) {
+      m_incPathsList.push_back(incpath);
+    }
```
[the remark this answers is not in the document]
brondani: @slhultgren In your example both de-duplications result in `IncludeA` as first element and `IncludeB` as second. You are telling me you would need `IncludeA` again even if it is already there, for what reason? Is it to solve a sort of circular dependency? Could you point to a practical project?

tools/buildmgr/cbuild/include/CbuildModel.h
```diff
@@ -380,7 +380,7 @@ class CbuildModel {
   std::map<std::string, std::list<std::string>>     m_cSourceFiles;
   std::map<std::string, std::list<std::string>>     m_cxxSourceFiles;
   std::map<std::string, std::list<std::string>>     m_asmSourceFiles;
-  std::set<std::string>                             m_includePaths;
+  std::vector<std::string>                          m_includePaths;
   std::set<std::string>                             m_packs;
   std::string                                       m_linkerScript;
   std::set<std::string>                             m_libraries;
```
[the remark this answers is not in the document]
slhultgren: Agreed, will update later today hopefully

PR #48, opened 2021-10-26
[packgen] Update CMake target's sources handling

brondani:
Avoid adding a directory as a source file into a pack component.
Extend test case.
Update documentation using inclusive language.


jkrech (approved):
Looks good to me.

PR #52, opened 2021-11-09
Added coverage workflow for packchk (#45)

soumeh01:


.github/workflows/packchk.yml
```diff
@@ -131,12 +131,13 @@ jobs:
 
       - name: Generate coverage report
         run: |
-          gcovr -r ../tools/packchk/src/ -b --xml -o coverage.xml ./tools/packchk/
```
JonatanAntoni: Ah, well. Using gcovr with -b causes branch coverage information to be emmitted.

JonatanAntoni (dismissed):

.github/workflows/packchk.yml
soumeh01: by default, gcovr emits information about branch coverage. It can't be excluded.
-b option is just to summarize the information on number of branches taken and the branches that were not completely covered.
 https://gcovr.com/en/stable/guide.html#options

JonatanAntoni (approved):

Issue #54, opened 2021-11-11
intdir / outdir folders are created in the wrong folder


slhultgren:
If --intdir=... / --outdir=... are not specified on the command line to cbuildgen, these folders will be created in the parent folder instead.

The reason for this is because the m_projectDir does not contain a trailing slash, which means that when concatenating it with a path, the result will be missing a separator for the last segment.

Steps to reproduce:
* Create a test.cprj that specifies 
```xml
<output intdir="int" outdir="out" ... />
```
* Run `.../test$ cbuildgen test.cprj cmake`
* Observe that `../testint` and `../testout` was created instead of `../test/int` and `../test/out`

brondani:
Hi @slhultgren, thanks for reporting this!

The issue seems to be a side effect of the `std::experimental::filesystem::path::remove_filename` function that removes the last directory separator, while the `std::filesystem::path::remove_filename` preserves it.
Can you please confirm that you are building with `std::experimental::filesystem`?
`std::experimental::filesystem` is deprecated is will be removed by most compiler vendors.

If we still need to support the experimental version an alternative to your suggestion could be to replace string concatenations by `std::filesystem::path::append` [operators](https://en.cppreference.com/w/cpp/filesystem/path/append) that should properly handle the directory separator.
What do you think?

slhultgren:
@brondani thanks, good catch, it does indeed seem to be an issue with the compiler itself.

I was originally building with "-G "Visual Studio 15 2017"", which resulted in "build engine 15.9" and this shows the issue.

I upgraded to "-G "Visual Studio 16 2019"" which has "build engine 16.11" and seems to solve the issue without the PR, so I closed my PR as invalid.

Maybe the README.md should update to only have "Visual Studio 2019"?

brondani:
> Maybe the README.md should update to only have "Visual Studio 2019"?

Agreed
https://github.com/Open-CMSIS-Pack/devtools/pull/56

PR #55, opened 2021-11-11
Ensure that projectDir path ends with '/'.

slhultgren:
Fixes issue #54 where folder paths where concatenated badly when
--intdir/--outdir are not specified on the command-line

Also applied helper function to intdir/outdir/workingdir

Signed-off-by: Samuel HULTGREN <samuel.hultgren@st.com>


slhultgren:
Invalid, seems to be an issue with Visual Studio 2017. Upgrading to build with Visual Studio 2019 solves the issue.

PR #56, opened 2021-11-12
Update README.md

brondani:
https://github.com/Open-CMSIS-Pack/devtools/issues/54


jkrech (approved):
I was hoping that we could continue supporting 2017 as well but it looks like 2019 has some fixes we want to rely on.

PR #62, opened 2021-11-24
Fix CPackGetTests issues

brondani:
- Fix windows unzip command line for preventing local environment failings
- Move `PackAddAlreadyInstalledPackTest` into `PackAddPackInstallationTest` because it was not self-contained and it was failing depending on the test order.
- Update `PackAddValidFileArgTest` to download smaller packs
- Fix typos

### DIFF
--- a/tools/buildmgr/cbuildgen/installer/create_installer.sh
+++ b/tools/buildmgr/cbuildgen/installer/create_installer.sh
@@ -73,7 +73,7 @@ cp ../../docs/LICENSE.txt ${distdir}
 # Get cpackget
 cpackget_version="0.2.0"
 cpackget_base=https://github.com/Open-CMSIS-Pack/cpackget/releases/download/v${cpackget_version}/cpackget_${cpackget_version}
-curl --retry 3 -L ${cpackget_base}_windows_amd64.zip   -o temp.zip && unzip -p temp.zip '*cpackget.exe' > ${distdir}/bin/cpackget.exe && rm temp.zip
+curl --retry 3 -L ${cpackget_base}_windows_amd64.zip   -o temp.zip && unzip -p temp.zip '*/cpackget.exe' > ${distdir}/bin/cpackget.exe && rm temp.zip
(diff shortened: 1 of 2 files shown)


tools/buildmgr/cbuildgen/installer/create_installer.sh
```diff
@@ -73,7 +73,7 @@ cp ../../docs/LICENSE.txt ${distdir}
 # Get cpackget
 cpackget_version="0.2.0"
 cpackget_base=https://github.com/Open-CMSIS-Pack/cpackget/releases/download/v${cpackget_version}/cpackget_${cpackget_version}
-curl --retry 3 -L ${cpackget_base}_windows_amd64.zip   -o temp.zip && unzip -p temp.zip '*cpackget.exe' > ${distdir}/bin/cpackget.exe && rm temp.zip
+curl --retry 3 -L ${cpackget_base}_windows_amd64.zip   -o temp.zip && unzip -p temp.zip '*/cpackget.exe' > ${distdir}/bin/cpackget.exe && rm temp.zip
```
chaws, line 76: Didn't `*` match `/` ? I may have overlooked this case

chaws (approved):
LGTM!

chaws:
@brondani Thanks for the fixes.
I was thinking about bringing those tests over to cpackget's repository and remove them from here. What do you think of it?

tools/buildmgr/cbuildgen/installer/create_installer.sh
brondani: In my local windows environment with unzip 6.00 I got the following error:
```
$ unzip -p temp.zip '*cpackget.exe'
caution: filename not matched:  *cpackget.exe
```

brondani:
> @brondani Thanks for the fixes. I was thinking about bringing those tests over to cpackget's repository and remove them from here. What do you think of it?

@chaws I agree, test cases that are specific for cpackget can/should live in the cpackget repo.

PR #65, opened 2021-12-07
Allow the "RTE" folder name to be configurable

slhultgren:
Adds optional rtedir attribute to /cprj/target/output element

e.g.

```xml
<output ... rtedir="CustomRTEFolder" />
```

Modified RteTestM3 project to use rtedir="CONFIG_FOLDER".
Renamed RTE to CONFIG_FOLDER in the test structure to reflect this

Builds on proposal from https://github.com/Open-CMSIS-Pack/Open-CMSIS-Pack/issues/30#issuecomment-935589272


libs/rtemodel/include/CprjFile.h
```diff
@@ -319,6 +320,7 @@ class CprjFile : public RtePackage
   */
   RteItem* GetProjectInfo() const;
 
+  const std::optional<std::string> GetRteFolder() const;
```
edriouk, line 323: Which advantage optional<> brings here to compare to string?

libs/rtemodel/include/RteProject.h
```diff
@@ -760,6 +767,7 @@ class RteProject : public RteItem
   std::map<std::string, RteModel*> m_targetModels; // filtered models for each target
   std::map<int, std::string> m_targetIDs;
   std::string m_sActiveTarget;
+  std::optional<std::string> m_rteFolder;
```
edriouk, line 770: again, what kind of advantage optional brings?

edriouk, line 770: Do we support one RTE folder per project, not per target?

libs/rtemodel/include/RteTarget.h
```diff
@@ -46,15 +46,17 @@ struct RteFileInfo
   /**
    * @brief compare file version of the given target and the instance
    * @param targetName target name
+   * @param rteFolder the "RTE" folder path used for placing files
    * @return 0 if both file versions are same, > 0 if file version of the given target is newer, otherwise < 0
   */
-  int HasNewVersion(const std::string& targetName) const;
+  int HasNewVersion(const std::string& targetName, const std::string& rteFolder) const;
```
edriouk, line 52: target knows its project, no need to pass en extra argument

libs/rtemodel/include/RteTarget.h
```diff
@@ -46,15 +46,17 @@ struct RteFileInfo
...
 
   /**
    * @brief compare file version of this instance with the ones of other targets
+   * @param rteFolder the "RTE" folder path used for placing files
    * @return 0 if equal to all others target, > 0 if file version of any other target is newer, otherwise < 0
   */
-  int HasNewVersion() const;
+  int HasNewVersion(const std::string& rteFolder) const;
```
edriouk, line 59: target knows its project, no need to pass en extra argument

libs/rtemodel/include/RteTarget.h
```diff
@@ -219,9 +221,10 @@ class RteTarget : public RteItem
    * @brief determine file given by instances of type RteFileInstance and RteComponent
    * @param fi given pointer of type RteFileInstance
    * @param c given pointer of type RteComponent
+   * @param rteFolder the "RTE" folder path used for placing files
    * @return pointer of type RteFile
   */
-  RteFile* GetFile(const RteFileInstance* fi, RteComponent* c) const;
+  RteFile* GetFile(const RteFileInstance* fi, RteComponent* c, const std::string& rteFolder) const;
```
edriouk, line 227: target knows its project, no need to pass en extra argument

libs/rtemodel/include/RteTarget.h
```diff
@@ -540,9 +543,10 @@ class RteTarget : public RteItem
 
   /**
    * @brief collect settings of given component instance
+   * @param rteFolder the "RTE" folder path used for placing files
    * @param ci pointer to RteComponentInstance object
   */
-  void CollectComponentSettings(RteComponentInstance* ci);
+  void CollectComponentSettings(RteComponentInstance* ci, const std::string& rteFolder);
```
edriouk, line 549: target knows its project, no need to pass en extra argument

libs/rtemodel/src/RteProject.cpp
```diff
@@ -103,6 +106,13 @@ RteProject* RteProject::GetProject() const
   return const_cast<RteProject*>(this);
 }
 
+const std::string& RteProject::GetRteFolder() const {
+  if (m_rteFolder.has_value()) {
```
edriouk, line 110: if( m_rteFolder.empty()) {
return DEFAULT_RTE_FOLDER;
}

libs/rtemodel/include/RteProject.h
```diff
@@ -149,6 +150,10 @@ class RteProject : public RteItem
   */
   void SetName(const std::string& name) { m_ID = name; }
 
+
+  void SetRteFolder(std::optional<std::string> rteFolder) { m_rteFolder = rteFolder; }
```
edriouk, line 154: Validation is missing!
the supplied string should : 
- be a valid folder name
- be different from folders mentioned in <files> section of the *.cprj file
Besides what should be done if project is already initialized? Copy content?

test/projects/RteTestM3/RteTestM3.cprj
```diff
@@ -31,7 +31,7 @@
   </compilers>
 
   <target Bname="RteTest Dummy board" Bvendor="Keil" Bversion="1.2.3" Ddsp="NO_DSP" Dendian="Little-endian" Dfpu="NO_FPU" Dmve="NO_MVE" Dname="RteTest_ARMCM3" Dsecure="Non-secure" Dtz="NO_TZ" Dvendor="ARM:82">
-    <output intdir="./" name="RteTestM3" outdir="./" type="exe"/>
+    <output intdir="./" name="RteTestM3" outdir="./" rtedir="CONFIG_FOLDER" type="exe"/>
```
edriouk, line 34: This test case is used in several places. It should not be modified.
Instead  an additional project should be used that specifies an RTE directory.
The corresponding config files should not be renamed either

edriouk (changes requested):

libs/rtemodel/test/src/RteModelTest.cpp
```diff
@@ -53,9 +53,10 @@ const string prjsDir = "RteModelTestProjects";
 const string RteTestM3 = "/RteTestM3";
 const string RteTestM3_cprj = prjsDir + RteTestM3 + "/RteTestM3.cprj";
 
-const string preIncComp = prjsDir + RteTestM3 + "/RTE/_RteTestM3/Pre_Include__RteTest_ComponentLevel.h";
-const string preIncGlob = prjsDir + RteTestM3 + "/RTE/_RteTestM3/Pre_Include_Global.h";
-const string rteComp = prjsDir + RteTestM3 + "/RTE/_RteTestM3/RTE_Components.h";
+const string RTE_FOLDER = "CONFIG_FOLDER";
```
edriouk, line 56: Instead of modifying an existing test, pleas add a new one

edriouk (changes requested):
The most important problem is that test is modified. 
Please leave the existing test as is (uses default RTE directory),
 but add a new one with custom RTE directory

libs/rtemodel/include/CprjFile.h
slhultgren: Maybe it is mostly personal preference, but for me the benefit of using optional is that it reduces the mental load on the programmer since I now don't have to remember when empty_string is considered either as OK, or usable value.
I can trust the created optional to "remember" this for me.

At the time when we create the optional, that is when we should take the decision if empty_string is an allowed value or not. Each user of the value should not have to remember this, but again, may be a personal preference thing.

slhultgren:
> The most important problem is that test is modified. Please leave the existing test as is (uses default RTE directory), but add a new one with custom RTE directory

Thanks for the review @edriouk I'll look into addressing the new test case, and avoiding the extra parameter for RteTarget hopefully today

libs/rtemodel/include/CprjFile.h
edriouk: Agreed, this might have sense

jkrech:
@slhultgren I would like to put this enhancement/feature on to the agenda for our next project meeting on Dec. 14th 2021. The reason I do not take this change lightly is the fact that new projects with a custom "rte" folder are no longer compatible with today's projects and we need to understand how a migration path should/would look like in case we agree we need one. In case there is further extensions that you have in the works or planned already it maybe good to learn about that, too.
Does that work for you?

libs/rtemodel/include/RteTarget.h
slhultgren: I was a bit worried here since the GetProject() may return nullptr sometimes. But looking through the current code it seems an RteTarget is always created with a parent that is at least an RteProject, so GetProject() should always return something in the current code.

If we start relying on this behaviour, perhaps it would make sense to change https://github.com/Open-CMSIS-Pack/devtools/blob/main/libs/rtemodel/include/RteTarget.h#L77 from `RteItem* parent` to `RteProject* parent` to avoid creating a "bad" RteTarget object in the future that would not have any associated project? What do you think?

libs/rtemodel/include/RteProject.h
slhultgren: Agreed, validation/sanitation should be added, looking at outdir is handled we should:
* Treat empty string as invalid
* Convert \ to /

Sorry, which section of folders do you refer to in the .cprj? but yes it could make sense to deny specifying rtedir == outdir || rtedir == intdir.

For me, as I update my cprj and change/set the value for the rtedir="..." I would at the same time manually also move the current RTE folder to the new name. Then the next time I run cbuildgen it would find an already initialized RTE folder (except with the new name)

So in my view cbuildgen does not need to remember the old name to try do something special in the case rtedir="..." changes.

libs/rtemodel/include/RteTarget.h
edriouk: We might change the signature to pass RteProject* has a parent. However it will be anyway stored as RteItem* in the base class.
 GetProject() may return nullptr in packchk executable:
`CheckConditions.cpp(23):m_target` = new RteTarget(nullptr, m_filteredModel, "CondTest", EMPTY_MAP);`

Anyway I would not change that in this PR. 
We already have some ideas on refactoring and improvement in the RTE. 
We can consider your proposal as well.

LMESTM:
@jkrech  should we follow-up on this item ?  
I think it is a good example of how "external contributors" can collaborate on the project.

jkrech (approved):
Sorry for this PR sitting here for so long. Even though we have not finalized the folders in the context of packmgr I accept this PR adding the ability to configure the name of the folder containing configuration files in the cprj file. If not specified "RTE" (runtime environment) is the default.

jeromecoutant:
Maybe it should be reflected in #77?

PR #66, opened 2021-12-07
[packchk] Workflow to publish test results (#59)

spcaipers-arm:


spcaipers-arm:
After this PR is merged, the Publish Test Reports will kick in.
This is an example how that will work:
https://github.com/spcaipers-arm/devtools/pull/4

Bottom Line: Once packchk (our first implementation) workflow is done, the "Publish Test Results" workflow will be started and it will store back the Test Results. So, it is independent for the PR workflow. This respects the GitHub Security Policy:
https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows
https://github.com/EnricoMi/publish-unit-test-result-action/blob/v1.20/README.md#support-fork-repositories-and-dependabot-branches

FYI: @soumeh01 .

JonatanAntoni (approved):

PR #69, opened 2021-12-10
[buildmgr] Fix EscapeQuotes function to handle multiple escapes (#67)

brondani:
This function is used when inserting toolchain flags into cmake variables `LD_FLAGS_GLOBAL` as well as `AS_FLAGS_*`, `CC_FLAGS_*` and `CXX_FLAGS_*` at all levels (global, groups, files).


jkrech (approved):
Looks good to me.

Issue #63, opened 2021-11-24
Diagram for loading project files is low resolution


benmcmorran:
[This diagram](https://github.com/Open-CMSIS-Pack/devtools/blob/main/libs/rtemodel/docs/ProjectLoad.png) is very low resolution and impossible to read.

jkrech:
Thanks for raising this. Let me just check whether this is better before getting the repo updated:

![ProjectLoad](https://user-images.githubusercontent.com/8290187/143445547-be11ea64-65cd-4aba-8cd2-e5d820a3b4d3.png)

benmcmorran:
That looks much better. Thanks a lot!

PR #70, opened 2021-12-16
Replace images with higher resolution #63

jkrech:


JonatanAntoni (approved):
LGTM.
Proposal: The images look like vector graphics. Where are the sources of them? I.e., without the source nobody else can modify/maintain the images. And for vector graphics we could store the images in SVG format instead of PNG to be independent of any specific resolution.

PR #72, opened 2021-12-16
Publish tests fix

spcaipers-arm:
This patch aims to fix Publish Tests on PRs.

This only affects packchk. There is no support for other tools yet.

Fixed:
- Avoid running it after the PR is merged on `main`.
- Using the correct path to look for the test result files.


spcaipers-arm:
The "Unit Test Results" will be fixed once this commit is merged.
Tested on: https://github.com/spcaipers-arm/devtools/pull/4

JonatanAntoni (approved):

PR #73, opened 2021-12-22
[projmgr] Handle local packs loading

grasci-arm:
[rtemodel] Extend and rework local packs discovering

### DIFF
--- a/tools/projmgr/src/ProjMgrKernel.cpp
+++ b/tools/projmgr/src/ProjMgrKernel.cpp
@@ -52,6 +52,18 @@ XMLTree* ProjMgrKernel::CreateXmlTree() const
 bool ProjMgrKernel::GetInstalledPacks(list<RtePackage*>& packs) {
   list<string> pdscFiles;
   RteFsUtils::GetPackageDescriptionFiles(pdscFiles, theProjMgrKernel->GetCmsisPackRoot(), 3);
+
+  // Find pdsc files from local repository
+  list<string> localPdscUrls;
+  if (!GetLocalPacksUrls(theProjMgrKernel->GetCmsisPackRoot(), localPdscUrls)) {
+    return false;
+  }
+  for (const auto& localPdscUrl : localPdscUrls) {
+    list<string> localPdscFiles;
+    RteFsUtils::GetPackageDescriptionFiles(localPdscFiles, localPdscUrl, 1);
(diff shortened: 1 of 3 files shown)


tools/projmgr/src/ProjMgrKernel.cpp
```diff
@@ -52,6 +52,18 @@ XMLTree* ProjMgrKernel::CreateXmlTree() const
 bool ProjMgrKernel::GetInstalledPacks(list<RtePackage*>& packs) {
   list<string> pdscFiles;
   RteFsUtils::GetPackageDescriptionFiles(pdscFiles, theProjMgrKernel->GetCmsisPackRoot(), 3);
+
+  // Find pdsc files from local repository
+  list<string> localPdscUrls;
+  if (!GetLocalPacksUrls(theProjMgrKernel->GetCmsisPackRoot(), localPdscUrls)) {
+    return false;
```
JonatanAntoni, line 59: Can we add a test case for this situation?

tools/projmgr/src/ProjMgrKernel.cpp
```diff
@@ -52,6 +52,18 @@ XMLTree* ProjMgrKernel::CreateXmlTree() const
...
+  }
+  for (const auto& localPdscUrl : localPdscUrls) {
```
JonatanAntoni, line 61: Can we add a test case for this situation?

JonatanAntoni (commented):
CodeCov points out some missing tests, can we add tests?

tools/projmgr/src/ProjMgrKernel.cpp
brondani: Yes, a local pack repository has to be created/emulated.

tools/projmgr/src/ProjMgrKernel.cpp
brondani: I believe this situation will happen only if the local pack repository is malformed, but it can be emulated and tested.

tools/projmgr/src/ProjMgrKernel.cpp
JonatanAntoni: Okay, this test case seems not to be reflected in the code coverage. We might need to check this at some point.

JonatanAntoni (approved):

PR #99, opened 2022-01-31
Fix auto-publish CMSIS-Build documentation

JonatanAntoni:


.github/workflows/buildmgr.yml
```diff
@@ -165,7 +166,9 @@ jobs:
         run: |
           VERSION=$(git show-ref ${{ github.ref }} | awk -F/ '{print $NF}')
           rm -rf buildmgr/${VERSION}
+          rm -f buildmgr/latest
           mkdir -p buildmgr/${VERSION}
+          ln -s ${VERSION}/ buildmgr/latest
```
brondani, line 171: I guess it should be:
```suggestion
          ln -s buildmgr/${VERSION}/ buildmgr/latest
```

JonatanAntoni: I think not, the target of a symlink needs to be relative to the link, not tu the current working directory. Please could you double check that e.g., in your local Linux env?

brondani: I have double checked, you are absolutely right.

brondani (approved):

PR #137, opened 2022-02-16
Fix ERROR M310 when directory name starts with .

### LINKED ISSUE #110: packchk : wrong ERROR M310 ?
Hi

It seems that function that is checking file name doesn't work when path name includes "." ?

```
*** ERROR M310: C:/xxx/STMicroelectronics.STM32U5xx_Drivers.pdsc (Line 597) 
  Filename mismatch (case sensitive):
  PDSC name : 'config/ReferenceFiles/CRC/stm32_crc_template.h'
  Filename  : '.config/ReferenceFiles/CRC/stm32_crc_template.h'
```

Here is STMicroelectronics.STM32U5xx_Drivers.pdsc (Line 597):
```
        <file category="other" name=".config/ReferenceFiles/CRC/stm32_crc_template.h"/>
```

jeromecoutant:
Fix #110 

See https://github.com/Open-CMSIS-Pack/devtools/issues/110#issuecomment-1033913423

@thorstendb-ARM


jeromecoutant:
> Merging [#137] will **decrease** coverage by `13.87%`.

I think there is an issue in Codecov tool....

thorstendb-ARM (dismissed):

JonatanAntoni:
> I think there is an issue in Codecov tool....
Yes, the CodeCoverage seems to be not properly configured. We need to check.

Can we add some tests to?
 * Reproduce the original issue.
 * Verify the fix does the job.

jeromecoutant:
> Can we add some tests for this function?

I think it's done :-)
 
>     Reproduce the original issue.

With only the commit with the new test:
```
$ cmake --build . --target packchk --target PackChkIntegTests

$ ctest -C Debug -R PackChkIntegTests

Test project C:/github/OpenCMSISPack/devtools/build
    Start 13: PackChkIntegTests
1/1 Test #13: PackChkIntegTests ................***Failed    1.27 sec

[  FAILED  ] PackChkIntegTests.CheckPackWithDot (7 ms)
```

>     Verify the fix does the job.

With the patch:
```
$ cmake --build . --target packchk --target PackChkIntegTests

$ ctest -C Debug -R PackChkIntegTests
1/1 Test #13: PackChkIntegTests ................   Passed    1.45 sec
```

JonatanAntoni:
Looks promising. I think we could add a unit test in addition to the integration test.

tools/packchk/test/data/TestPackDot/TestVendor.TestPackDot.pdsc
```diff
@@ -0,0 +1,36 @@
...
+      <description>Test Device</description>
+      <require Dvendor="ARM:82"/>
+    </condition>
+  </conditions>
+
+  <components>
+    <component Cclass="TestClass" Cgroup="TestGlobal" Cversion="1.0.0" condition="Test_Condition">
+      <description>TestGlobal</description>
+      <files>
+        <file category="source" name=".Files/test1.c"/>
```
JonatanAntoni, line 30: What are we trying to test here? Filenames starting with a `.` should be legal. At least on unix-like systems such files/directories are considered "hidden".
Didn't we discuss (and test) for names starting with `./`? I.e., the unix-like shortcut for "current working directory"?

JonatanAntoni (changes requested):
The coverage reports turns out that the code range under discussion is not properly hit by the provided tests.

tools/packchk/test/data/TestPackDot/TestVendor.TestPackDot.pdsc
JonatanAntoni: Okay, so this test actually does not verify the removal of `./` prefixes, intentionally. So we definitively need to add some more tests, probably unit tests, to really cover the implementation.

jeromecoutant:
> So we definitively need to add some more tests, probably unit tests, to really cover the implementation

Any update ?
Thx

fred-r:
What can we do about the codecov issue blocking the merge ?

jeromecoutant:
@JonatanAntoni any update ?
Thx

JonatanAntoni:
@jeromecoutant,

We need to provide more tests to be sure the function does what is expected.
Please rebase your patch and add some tests which cover the code under modification.

Cheers,
Jonatan

jeromecoutant:
> Please rebase your patch and add some tests which cover the code under modification.

Done

JonatanAntoni:
@jeromecoutant, after merging #262 please consider rebasing this PR to keep contributing your integration test. Actually, we should already have everything covered with the unit tests but the additional integ test would not harm.

jeromecoutant:
@JonatanAntoni ?

JonatanAntoni (approved):

Issue #88, opened 2022-01-25
[projmgr] Open door to vendor specific project(s)


VGRSTM:
I foresee csolution.yml like the good way to describe as end user a complex application relying on multiple projects like https://github.com/Open-CMSIS-Pack/devtools/blob/main/tools/projmgr/test/data/TrustZoneSolutionLayers/TrustZone.csolution.yml.
I'm discussing here only about relationship between projects description.

We have to anticipate a vendor may introduce some project(s) which is(are) not fully Open-CMSIS compliant. Open-CMSIS compliant means to get bare minimum & meaningful required cproject.yml content.
A vendor may promote as part of application project dependencies (because useful to get dependencies described somewhere) some contributions which are not component aware ones.

I would push so for a new key if cproject.yml allowing [projmgr] to ignore it as part of common processing.

VGRSTM:
May propose following schema [common.schema.json](https://github.com/Open-CMSIS-Pack/devtools/blob/main/tools/projmgr/schemas/common.schema.json)

![isopencmsis](https://user-images.githubusercontent.com/71082832/155322797-0f9b2cc1-c76f-461a-8e91-02fa2e03deb3.PNG)

If boolean set to `false`, let's Open-CMSIS devtools just ignore such cproject file as part of processing (like csolution processing)

VGRSTM:
Not so much comment on such proposal. Let's say we can rely on [[projmgr] Open door to vendor specific data addon](https://github.com/Open-CMSIS-Pack/devtools/issues/83) despite possibly having then at many place if csolution some project dependencies description. Let's come back on such later on if at end user pain at a point.

Closing issue in sync.

Issue #83, opened 2022-01-24
[projmgr] Open door to vendor specific data (& processing) addon


VGRSTM:
As common good practice we have to find a way to allow vendor specific data addon to project manager metadata (*.yml files).
See https://github.com/Open-CMSIS-Pack/devtools/pull/77/files#r787702533

Intention is to get Open-CMSIS project manager as smart as possible but just incase we cannot anticipate all vendor specific corner cases / integration context etc. 
Even thinking continuous improvement of devtools, sounds nice to consider a contributor may first Thanks some official backdoor validate (possibly in production) its future addon to main solution.

I guess if we want to ensure a robust solution, no choice, *.yml stuff has to be validated up front against a schema. So I do not foresee any chance to support custom material addon to common basis. 

Do we have then to add as an option which is part of main schema some include like statement ? Such include statement may allow to get a vendor specific *.\<vendor\>.yml close to and referenced by related main *.yml file.

`solution:`
`- project: ./SWProject_1/SWProject_1.cproject.yml`
`- project: ./SWProject_2/SWProject_2.cproject.yml`
**`vendor-extension: ./.extended.csolution.yml`**

Today produces `.csolution.yml:4:1 - warning projmgr: key 'vendor-extension' was not recognized`
Good is no blocking process but long run I would prefer to block it (more robust) preventing at same time spurious trace.
Primary *.yml has to be validated strictly against an Open-CMSIS schema. Vendor specific part if under vendor responsibility.

ReinhardKeil:
I would like to understand this proposal.  Is it basically a suggestion to add in each *.yml file the possibility to add a section "vendor-extension:" as shown below.  This would of course apply to *.csolution.yml, *.cproject.yml, and *.clayer.yml files.

```yml
solution:
  compiler: AC6

  target-types:
    - type: Board
      board: NUCLEO-L552ZE-Q

    - type: Production-HW
      device: STM32L552XY         # specifies device

    - type: Virtual
      board: VHT-Corstone-300     # Virtual Hardware platform (appears as board)
      
  build-types:
    - type: Debug
      optimize: debug
      debug: on

  vendor-extension:                   # start of vendor extensions
                                                  # a section that is not checked by the tool
```

VGRSTM:
@ReinhardKeil yes purpose is to add "vendor-extension:". My proposal including a vendor specific file is way I foresee to support addon without breaking schema check.

ReinhardKeil:
[...]

PR #163, opened 2022-02-25
[projmgr] Open door to vendor specific data addon adding open-cmsis scope

VGRSTM:
Proposal in sync. #83
Such will preserve current OpenCMSIS proposal but adding `open-cmsis` scope.
Is allowing integrator to promote its own schema on top of.
Introducing scope may helps clarity I guess plus may be useful if #88 `devtools` having only to consider if `open-cmsis` scope is available or not while iterating on csolution.yml referenced *.cproject.yml file(s)


brondani:
Closing in benefit of https://github.com/Open-CMSIS-Pack/devtools/pull/162

PR #234, opened 2022-03-23
[projmgr] Open door to vendor specific data addon: schema polishing

VGRSTM:
Previous JSON schema was valid (https://www.jsonschemavalidator.net/) but possibly not best.
Removes useless lines preventing https://quicktype.io/ to run.
`"additionalProperties": false` is part already of `./common.schema.json`

### DIFF
--- a/tools/projmgr/schemas/cdefault.schema.json
+++ b/tools/projmgr/schemas/cdefault.schema.json
@@ -6,8 +6,7 @@
   "properties": {
     "default": {
       "$ref": "./common.schema.json#/definitions/DefaultDescType"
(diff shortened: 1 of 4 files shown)


brondani (commented):
@VGRSTM Can you please apply the same change to the [cdefault.schema.json](https://github.com/VGRSTM/devtools/blob/vgrstm-issue83_improvement/tools/projmgr/schemas/cdefault.schema.json) file?

brondani (approved):
@VGRSTM Thanks for this contribution!
LGTM

PR #242, opened 2022-03-29
[buildmgr] Use generated Translation controls

grasci-arm:


brondani (approved):

jeromecoutant:
Hi
IAR cmake file has not been updated with this PR ?

@fred-r
@tarek-bochkati

brondani:
@jeromecoutant Thanks for pointing this out, this shall be addressed by IAR cmake maintainer @tarek-bochkati

brondani:
Hi @tarek-bochkati

Translation Control is a generic name for the several compile options as listed here:
https://github.com/Open-CMSIS-Pack/devtools/blob/main/tools/projmgr/docs/Manual/YML-Format.md#translation-control

In this PR we have extended the support for `includes`/`defines` that we had only at target level also for component, group and file level (and symmetrically also `excludes`/`undefines`).

The rework to be done in the `IAR.<version>.cmake` file should be similar to what is implemented here for GCC:
https://github.com/Open-CMSIS-Pack/devtools/blob/main/tools/buildmgr/cbuildgen/config/GCC.10.2.1.cmake#L106-L132

In others words the CMake function `cbuild_set_defines` should be added replacing the `defines` handling. This new function is used by the new generated CMakeLists to set the defines at a finer granularity when needed (file level).

Many thanks for your help!

tarek-bochkati:
Thanks @brondani 
I have added a tested the requested change.
@see PR #325

PR #247, opened 2022-03-29
[projmgr] Update schema versions

grasci-arm:


brondani (approved):

jeromecoutant:
Hi
Seems it is a heavy process to update 102 files each times a new version is released... :-(

Issue #221, opened 2022-03-16
[csolution] [cbuild] directory tree gets difficult to use


fred-r:
Let's consider a csolution with 1 board and 3 compilers:
```
$ tree
.
|-- README_OpenCMSIS_Project.md
|-- gpio-toggle.cproject.yml
`-- gpio-toggle.csolution.yml

0 directories, 3 files
```

Then I create the cprjs:
```
$ csolution -s gpio-toggle.csolution.yml convert
C:/Cube/stm32cube_fw_all/examples/hal/gpio/toggle/projects/gpio-toggle.AC6+B-U585I-IOT02A.cprj - info csolution: file generated successfully
C:/Cube/stm32cube_fw_all/examples/hal/gpio/toggle/projects/gpio-toggle.GCC+B-U585I-IOT02A.cprj - info csolution: file generated successfully
C:/Cube/stm32cube_fw_all/examples/hal/gpio/toggle/projects/gpio-toggle.IAR+B-U585I-IOT02A.cprj - info csolution: file generated successfully


$ tree
.
|-- README_OpenCMSIS_Project.md
|-- RTE
|   `-- Device
|       `-- STM32U585ZITx
|           |-- STM32U585xx_FLASH.ld
|           |-- stm32_external_env.h
|           |-- stm32_external_env.h@0.0.1
|           |-- stm32u585xx_flash.icf
|           |-- stm32u585xx_flash.sct
|           |-- system_stm32u5xx.c
|           `-- system_stm32u5xx.c@0.0.1
|-- gpio-toggle.AC6+B-U585I-IOT02A.cprj
|-- gpio-toggle.GCC+B-U585I-IOT02A.cprj
|-- gpio-toggle.IAR+B-U585I-IOT02A.cprj
|-- gpio-toggle.cproject.yml
`-- gpio-toggle.csolution.yml

3 directories, 13 files
```

Already at this stage it becomes pretty fuzzy with 1 RTE with 1 device but 3 CPRJs at the same level.

Now, if I build:
```
$ cbuild.sh gpio-toggle.AC6+B-U585I-IOT02A.cprj

$ tree
.
|-- README_OpenCMSIS_Project.md
|-- RTE
|   |-- Device
|   |   `-- STM32U585ZITx
|   |       |-- STM32U585xx_FLASH.ld
|   |       |-- stm32_external_env.h
|   |       |-- stm32u585xx_flash.icf
|   |       |-- stm32u585xx_flash.sct
[...]

PR #266, opened 2022-04-05
[csolution] [cbuild] directory tree gets difficult to use.

VGRSTM:
Add capability to define custom output directories.

### DIFF
--- a/tools/projmgr/schemas/common.schema.json
+++ b/tools/projmgr/schemas/common.schema.json
@@ -55,6 +55,16 @@
       "enum": [ "exe", "lib" ],
       "description": "Generate executable (default) or library (exe, lib)."
     },
+    "OutputDirectoriesType": {
+      "type": "object",
+      "properties": {
+        "rtedir":   { "type": "string" },
+        "intdir":   { "type": "string" },
+        "outdir":   { "type": "string" },
+        "cprjdir":  { "type": "string" }
+      },
+      "additionalProperties": false
+    },
     "DefinesType": {
       "type": "array",
       "description": "List of #define symbol settings for the code generation.",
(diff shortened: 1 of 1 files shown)


VGRSTM:
Address https://github.com/Open-CMSIS-Pack/devtools/issues/221 schema part

jkrech:
I think it is not wise to include 'rtedir' here. From my perspective the content of today's 'RTE' directory is part of the project source (it always needs to be committed to revision control system).
The other directories specify (reproducible) build information, intermediate build output and build artifacts.
In the context of layer handling as well depending on whether we plan to support multiple cprojects per single directory, we may need further review of the RTE directory.

VGRSTM:
@jkrech not sure to understand your trouble here if RTE

Default can be:
> output-dirs:
 >   rtedir: ./RTE
 
 Renaming `RTE` folder is part already of [Allow the "RTE" folder name to be configurable](https://github.com/Open-CMSIS-Pack/devtools/pull/65). Such contribution was handling `cprj` only. Now we have `yml` so just promoting same here.

ReinhardKeil:
Maybe we can extend this https://github.com/Open-CMSIS-Pack/devtools/issues/314#issuecomment-1118243602

brondani (approved):

VGRSTM:
@brondani Thanks for such contribution !
I've got a try relying on some local build. Works as expected on my local example project. I've been got thru project manager and build manager workflow with some success.

image

I'm getting as expected new defaults in sync. #314 
Only comment is `cproject.yml` `output-dirs` settings if some are not applied per my trials.

brondani:
> Only comment is cproject.yml output-dirs settings if some are not applied per my trials.

Thanks for the feedback @VGRSTM
As I commented [here](https://github.com/Open-CMSIS-Pack/devtools/issues/314#issuecomment-1174862661) it is purposely implemented only at csolution level, so it is not handled yet in `cproject.yml`. Let's get it consolidated before extending it.

Issue #224, opened 2022-03-18
[cbuildgen] Add C and C++ linker flags support


VGRSTM:
As part of [CPRJ.xsd](https://github.com/Open-CMSIS-Pack/devtools/blob/main/tools/buildmgr/cbuildgen/config/CPRJ.xsd) we have 
>  <xs:complexType name="TargetType">
    <xs:choice maxOccurs="unbounded">
      <\!-- Build options -->
      <xs:element name="output"    type="OutputType" />
      <\!-- Linker command-line -->
      <xs:element name="ldflags"   type="LinkerFlagsType" />
      <\!-- C-Compiler command-line -->
      <xs:element name="cflags"    type="ToolOptionType" />
      <\!-- C++ Compiler command-line -->
      <xs:element name="cxxflags"  type="ToolOptionType" />
      <\!-- Archiver command-line -->
      <xs:element name="arflags"  type="ToolOptionType" />
      <\!-- Assembler command-line -->
      <xs:element name="asflags"   type="AsmOptionType" />
      <\!-- Include paths -->
      <xs:element name="includes"  type="xs:string" minOccurs="0" />
      <\!-- Preprocessing definitions -->
      <xs:element name="defines"   type="xs:string" minOccurs="0" />
    </xs:choice>
    
C compiler flags are addressed Thanks `cflags` element.
C++ compiler flags are addressed Thanks `cxxflags` element.
Sounds weird to get only `ldflags` supporting both C **and** C++ ..

Would push to introduce `ldxxflags` element as addon to `ldflags`

brondani:
@VGRSTM Please note `ldflags` are not specific for one or another language.
Regardless of having mixed code, a single target will likely require a single linker.
Why would you use multiple sets of `ldflags` in a single `CPRJ`?

VGRSTM:
Linker flags may vary if C or C++ project at least if GCC, let's think about built-in libraries.
- C++ :  -Wl,--start-group -lc -lm -lstdc++ -lsupc++ -Wl,--end-group
- C:  -Wl,--start-group -lc -lm -Wl,--end-group

jkrech:
@VGRSTM I do agree that the linker flags will definitely depend on whether the objects you try to link are C++ or C. 
However in the scope of a CPRJ file there is only a single invocation of the linker and it can be either be linking with one or the other set of flags depending on the objects that get linked.
[...]

PR #302, opened 2022-04-28
C and C++ linker flags support addon - devtools issue #224

VGRSTM:
Serving https://github.com/Open-CMSIS-Pack/devtools/issues/224

### DIFF
--- a/tools/buildmgr/cbuildgen/config/CPRJ.xsd
+++ b/tools/buildmgr/cbuildgen/config/CPRJ.xsd
@@ -296,9 +296,11 @@
     <xs:choice maxOccurs="unbounded">
       <!-- Build options -->
       <xs:element name="output"    type="OutputType" />
-      <!-- Linker command-line -->
+      <!-- C Linker command-line -->
       <xs:element name="ldflags"   type="LinkerFlagsType" />
-      <!-- C-Compiler command-line -->
+      <!-- C++ Linker command-line -->
+      <xs:element name="ldxxflags"   type="LinkerFlagsType" />
+      <!-- C Compiler command-line -->
       <xs:element name="cflags"    type="ToolOptionType" />
       <!-- C++ Compiler command-line -->
       <xs:element name="cxxflags"  type="ToolOptionType" />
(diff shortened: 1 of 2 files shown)


VGRSTM:
@brondani  not sure to understand why ACI is failing here ... ?

brondani:
> @brondani not sure to understand why ACI is failing here ... ?

Test projects containing the `Link` keyword are failing because this PR is removing it from the schema.
For example here: https://github.com/Open-CMSIS-Pack/devtools/runs/6211054212?check_suite_focus=true#step:5:572

jkrech:
This is a backward incompatible change which would require to also include the source code changes in the tools as well as the changes in the tests as part of this pull request.
Can we consider a backward compatible change and deprecate the Link over time to not force examples to change immediately.

VGRSTM:
Ok my bad have not looked at details per OSes log is more explicit for sure.
I've got issue now. Thanks.

>This is a backward incompatible change

I agree so. We come back to past discussion about e/o May toolbox tag will it be 1.0.0 or 0.x.y ... sounds to me too early to promote 1.0.0 ... 0.x.y allows according to me backward compatibility breaks ...

ReinhardKeil:
Just to clearify expectations for May:  we want a stable release, but qualified as beta (V0.xx).  Currently it is alpha.  Beta does not give any backwards compatiblity.  V1.0 would do that - which I don't think we will reach until end of May.

VGRSTM:
@ReinhardKeil I'm perfectly fine with and better to get it clearly stated. Thanks for that.

@jkrech I foresee so no issue to 
>force examples to change immediately

Still @jkrech  my pain is I cannot commit shortly having bandwidth / capability to propagate such update to tools & test non regression suites despite I perfectly acknowledge your request here.

ReinhardKeil:
@VGRSTM this is quite old.  Is it still relevant?

VGRSTM:
@ReinhardKeil 
> @VGRSTM this is quite old. Is it still relevant?

Yes it is. According to me no update since related to this area. We still have need to explicit variations if link process depending if C or Cpp project.

ReinhardKeil:
Should be now covered under [misc](https://github.com/Open-CMSIS-Pack/devtools/blob/main/tools/projmgr/docs/Manual/YML-Input-Format.md#misc) with Link-C: and Link-CPP:

brondani:
This has been implemented in the PR https://github.com/Open-CMSIS-Pack/devtools/pull/492

PR #322, opened 2022-05-06
[buildmgr] Fix legacy assembly defines

grasci-arm:


jeromecoutant:
@tarek-bochkati for IAR

Issue #179, opened 2022-03-07
[projmgr] [buildmgr]  Opportunity to introduce toolchain-independent configuration using the most common generic configurations


VGRSTM:
Discussion initiated if https://github.com/Open-CMSIS-Pack/devtools/issues/120

https://github.com/Open-CMSIS-Pack/devtools/issues/120#issuecomment-1050612457
https://github.com/Open-CMSIS-Pack/devtools/issues/120#issuecomment-1050634268
https://github.com/Open-CMSIS-Pack/devtools/issues/120#issuecomment-1055457730

Sounds good to me to get dedicated issue to comment / amend on.

Discussion is about following yml schema contributions: https://github.com/Open-CMSIS-Pack/devtools/blob/main/tools/projmgr/schemas/common.schema.json
>    "OptimizeType": {
      "enum": [ "max", "size", "speed", "debug" ],
      "description": "Generic optimize levels (max, size, speed, debug) for code generation."
    },
    "DebugType": {
      "enum": [ "on", "off" ],
      "description": "Generic control for the generation of debug information (on, off)."
    },
    "WarningsType": {
      "enum": [ "no", "all", "Misra", "AC5-like" ],
      "description": "Control warnings (no, all, Misra, AC5-like)."
    },

ReinhardKeil:
The is inline with proposal #315

tarek-bochkati:
after the translation controls options addition (ie. optimize, debug, warnings),
I have started supporting them into `projmgr` and `buildmgr`
while doing that I had some notes which I need to share with you:
* `WarningsType` enum `AC5-like` is only valid for AC6 compiler,
   thus I suggest either to remove it from the standart or add a mention in the documentation saying it's for AC6 only
* `OptimizeType`
  * I can agree that `debug` means the lowed optimization possible + plus adding more debug info into the generate ELF file. 
    thus I'm wondering if it is worth replacing it by `none` and we can use the debug options to add debug information
    ```
    if optimize=none
        if debug=on
            use -Og
        else
            use -O0
        endif
    endif
    ```
  * `max` is not clear for me:
    * for AC6 it should be linked to LTO: but is it `-Omin` (minimal size) or `-Omax` (best performance)
    * what about AC5 does it implies using `Cross-Module Optimization` ?
      and the same question as for AC6, is it using `-Ospace -O3` or `-Otime -O3`
    * IAR does not have equivalent for `-Omax` optimization,
[...]

PR #415, opened 2022-07-25
ProfileType addon - devtools issue #179

VGRSTM:
Serving [comment](https://github.com/Open-CMSIS-Pack/devtools/issues/179#issuecomment-1193667376) if  [[projmgr] [buildmgr] Opportunity to introduce toolchain-independent configuration using the most common generic configurations](https://github.com/Open-CMSIS-Pack/devtools/issues/179)


ReinhardKeil (commented):
I think this pull request is somewhat outdated as Tarek was working on that.

VGRSTM:
> I think this pull request is somewhat outdated as Tarek was working on that.

Yes and no ... such `ProfileType` proposal is a way to get rid or addition possibly of toolchain abstractions supported today by `OptimizeType`, `DebugType`, `WarningType`. My 2 cents were and are I'm a bit doubtful today abstraction is going to be valuable enough. Having "only" boolean if  `DebugType`, `WarningType` and very limited flexibility if `OptimizeType` sounds fine if basic / common usage but.

jkrech:
@VGRSTM I would like to suggest to close this PR for now and work with the proposed separate controls
`debug`, `optimize`, `warning`
We should review the usage in the examples to see whether used patterns could be expressed more effectively with profiles.

ReinhardKeil:
@VGRSTM since the tool agnostic controls are working, is this pull request still needed?  I think it landed somewhat different?

VGRSTM:
> @VGRSTM since the tool agnostic controls are working, is this pull request still needed? I think it landed somewhat different?

Let's close for now. If at a time we come back to https://github.com/Open-CMSIS-Pack/devtools/pull/415#issuecomment-1243495874 we will have opportunity to find this material back.

PR #419, opened 2022-07-26
Add [abc] style wildcard to pack name convention

madchutney:
The `[abc]` style wildcard is already supported as it is a common implementation.

Updating the documentation to be consistent with https://open-cmsis-pack.github.io/Open-CMSIS-Pack-Spec/main/html/pdsc_conditions_pg.html#element_conditions


edriouk (approved):

brondani:
@madchutney Thanks for this contribution!

PR #426, opened 2022-08-02
Update code coverage steps

JonatanAntoni:
- Don't archive coverage as zip in zip
- Include lcov information in archive
- Add branch information to lcov


thorstendb-ARM (approved):
looks reasonable

PR #430, opened 2022-08-16
[projmgr] Reworked 'device not found' message (#278)

grasci-arm:
* [projmgr] Reworked 'device not found' message

* Resolved warning by removing escape character

### DIFF
--- a/tools/projmgr/src/ProjMgrWorker.cpp
+++ b/tools/projmgr/src/ProjMgrWorker.cpp
@@ -479,7 +479,9 @@ bool ProjMgrWorker::ProcessDevice(ContextItem& context) {
       }
     }
     if (!matchedDevice) {
-      ProjMgrLogger::Error("specified device '" + deviceItem.name + "' was not found");
+      string msg = "specified device '" + deviceItem.name + "' was not found among the installed packs.";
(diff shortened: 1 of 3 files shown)


tools/projmgr/src/ProjMgrWorker.cpp
```diff
@@ -479,7 +479,9 @@ bool ProjMgrWorker::ProcessDevice(ContextItem& context) {
       }
     }
     if (!matchedDevice) {
-      ProjMgrLogger::Error("specified device '" + deviceItem.name + "' was not found");
+      string msg = "specified device '" + deviceItem.name + "' was not found among the installed packs.";
+      msg += "\nuse \'cpackget\' utility to install software packs.\n  cpackget add Vendor.PackName --pack -root ./Path/Packs";
```
brondani, line 483: The cpackget `--pack-root` option should be corrected here and in the test cases:
```suggestion
      msg += "\nuse \'cpackget\' utility to install software packs.\n  cpackget add Vendor.PackName --pack-root ./Path/Packs";
```

brondani (changes requested):

brondani (approved):

PR #432, opened 2022-08-17
Make creation of RTE folder optional

### LINKED ISSUE #371: Optional RTE creation
To make the tools easier to use for testing and validation and to help the user control when and why the RTE structure is created/updated, it would be desirable to have a flag to disable/enable RTE file creation for both cbuildgen.exe and csolution.exe.

This would allow to generate CPRJ and/or CMakeLists.txt without creating the RTE structure.

Related to https://github.com/Open-CMSIS-Pack/devtools/issues/85

ErikLundinSTM:
This introduces the flag `--no-update-rte` for the `csolution convert` and `cbuildgen cmake` commands to skip creation/update of the RTE folder and files. It is useful when you just want to resolve the project but don't want the RTE files.

In addition, the commands `csolution list dependencies` and `csolution list generators` are fixed to never create the RTE folder.

Fixes #371

Prerequisite: #436

Signed-off-by: Erik LUNDIN <erik.lundin@st.com>


ErikLundinSTM:
One of the integration tests failed so I pushed fixed versions of the commits.

libs/rtemodel/include/RteKernel.h
```diff
@@ -127,18 +127,20 @@ class RteKernel
...
   /**
    * @brief initializes loaded Rte
    * @param cprjProject pointer to RteCprjProject
    * @param toolchain compiler toolchain to select, if empty non-ambiguous default is used
    * @param toolChainVersion, use default if non-ambiguous
+   * @param generateRteHeaders whether to generate RTE headers or not
    * @return true if successful
   */
-  bool InitializeCprj(RteCprjProject* cprjProject, const std::string& toolchain = RteUtils::EMPTY_STRING, const std::string& toolChainVersion = RteUtils::EMPTY_STRING);
+  bool InitializeCprj(RteCprjProject* cprjProject, const std::string& toolchain = RteUtils::EMPTY_STRING, const std::string& toolChainVersion = RteUtils::EMPTY_STRING, bool generateRteHeaders = true);
```
edriouk, line 143: RteKernel is a singleton. There is no need to pass a boolean flag all way down in the hierarchy.
Instead RteKernel can contain methods like SetGenerateRteHeaders(bool)  and IsGenerateRteHeaders()  
or/and  IsCopyRteFiles(),  SetCopyRteFiles();

edriouk (changes requested):
There are three problems with this PR that need to re-worked: 
1. Separate RteFsUtils change in a separate PR as it is not related to the rest.
2. Add RteKernel methods to check if RTE files need to be copied. This should replace passing  such flags all way down to the hierarchy.
3. Restructuring InitFileInstance() and adding WriteInstanceFiles() : all cses should be considered, not only loading cprj file

ErikLundinSTM:
> There are three problems with this PR that need to re-worked:

Thank you for your suggestions. Regarding passing the flag down the hierarchy, I agree that it felt somewhat strange to have to do that, and I'm looking into how to implement it using `RteKernel`.

libs/rtefsutils/include/RteFsUtils.h
```diff
@@ -248,9 +248,8 @@ class RteFsUtils
    * @brief normalize path by ensuring that it's absolute and canonical
    * @param path folder to be normalized
    * @param base root folder to form absolute
-   * @return true if path is normalized
   */
-  static bool NormalizePath(std::string& path, std::string base = "");
+  static void NormalizePath(std::string& path, std::string base = "");
```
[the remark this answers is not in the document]
ErikLundinSTM: This PR depends on this change, since evaluation will fail if the normalized paths don't exist when the RTE files not have been created. I can put this in a separate PR if you like, but then this PR will depend on that one.

libs/rtemodel/src/RteProject.cpp
```diff
@@ -449,14 +448,28 @@ void RteProject::InitFileInstance(RteFileInstance* fi, RteFile* f, int index, Rt
   fi->SetRemoved(false);
   string absPath = fi->GetAbsolutePath();
   bool bExists = RteFsUtils::Exists(absPath);
-  if (!bExists) {
-    UpdateFileInstance(fi, f, false, false);
-  } else {
+  if (bExists) {
     fi->AddAttribute("version", savedVersion, false);
   }
-  UpdateConfigFileBackups(fi, f);
```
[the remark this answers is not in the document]
ErikLundinSTM: > 3\. Restructuring InitFileInstance() and adding WriteInstanceFiles() : all cses should be considered, not only loading cprj file

Were there any specific cases you had in mind that wasn't covered?

libs/rtemodel/include/RteKernel.h
ErikLundinSTM: > 3\. Restructuring InitFileInstance() and adding WriteInstanceFiles() : all cses should be considered, not only loading cprj file

Were there any specific cases you had in mind that wasn't covered?

libs/rtemodel/src/RteProject.cpp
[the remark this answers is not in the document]
ErikLundinSTM: The call to `UpdateConfigFileBackups` creates files in the RTE folder, so that's why I break it out and put in the new `WriteInstanceFiles` that can be called separately based on whether RTE files should be updated or not.

libs/rtefsutils/include/RteFsUtils.h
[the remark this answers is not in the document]
ErikLundinSTM: New PR with this change submitted as #436

libs/rtemodel/include/RteKernel.h
edriouk: Unfortunately, other cases are outside this repository, but rely on current implementation :(
In particular, any call to RteProject::Update() should trigger file copy (if needed)
I do agree however that writing config files could be separated from updating internal structures.

ErikLundinSTM:
> 2\. Add RteKernel methods to check if RTE files need to be copied. This should replace passing  such flags all way down to the hierarchy.

I ended up adding methods to `RteModel` instead of `RteKernel`, since the flag needs to be checked also in `RteProject`, which doesn't have direct access to the kernel but to the global model.

brondani:
@ErikLundinSTM As discussed in the meeting today, let's make the inverted logic for `cbuildgen cmake`, in other words by default don't create/update RTE files but introduce an `--update-rte` flag.
Please investigate why [projmgr tests](https://github.com/Open-CMSIS-Pack/devtools/actions/runs/2911661520/jobs/4637285229) and [buildmgr tests](https://github.com/Open-CMSIS-Pack/devtools/actions/runs/2911661523/jobs/4637366078) are failing.
The tests must cover both behaviours (create/not create RTE) and all RTE content (config files, RTE_Components.h, pre-include files) for csolution and cbuildgen.
Feel free to contact us for any kind of support.
@edriouk Please write here your considerations so we can keep track, thanks!

edriouk:
I do agree with the idea that updating RTE folders can be made optional.
Implementation however should be backward compatible and not break the existing client implementations.
That  can be achieved by the following measures: 
1. The calls to methods that copies, creates or deletes files should deletes files in the folders should remain as they are.
2. The signatures of those methods should not be changed.
3. The methods should themselves query parent RteProject if to perform the actions. 
4. The RteProject can keep that flag in its attributes, e.g. as "update-rte" boolean attribute (default true). Setting that attribute can be performed with the existing methods. **NOTE**: cbuildgen must always explicitly set the attribute!

ErikLundinSTM:
@edriouk Thank you for your input. When saying that the methods that copy, create or delete files/folders should be left unchanged, do you refer to e.g. `RteProject::AddCprjComponents` and `RteProject::InitFileInstance`? They can probably be changed back to keep updating the files, based on the value of an attribute set on the project, but wouldn't that mean that we lose the ability to explicitly do all the file operations? One important idea behind this PR was to extract the RTE file operations from various places to be able to have better control of when it happens.

edriouk:
@ErikLundinSTM, yes, I would suggest changing RteProject::AddCprjComponents and RteProject::InitFileInstance back to their previous states to keep compatibility with the existing tools first. The most important aspect here is that  RteProject::InitFileInstance() gets eventually called from the following chain: 
```RteProject::Apply()->RteProject::Update()->AddComponentFiles()->... -> RteProject::InitFileInstance()```
That chain gets executed by loading a project and every time by adding/removing components. Thus not only via RteProject::AddCprjComponents().

The obvious place to centralize file operations is RteProject::Update().  I would prefer to do that consolidation in a separate PR.

**Comment by ErikLundinSTM** [comment]
I have modified the PR to not change the mentioned functions. All tests should pass now as well. Please check if there are more things you would like to have changed.

File: tools/buildmgr/cbuild/include/CbuildModel.h [hunk 4]
```diff
@@ -508,19 +508,19 @@ class CbuildModel {
...
-  bool EvalFile(RteItem* file, const std::string& group, const std::string& base, std::string& filepath);
-  bool EvalItem(RteItem* item, const std::string& groupName = std::string(), const std::string& groupLayer = std::string());
+  bool EvalRteSourceFiles(std::map<std::string, std::list<std::string>> &cSourceFiles, std::map<std::string, std::list<std::string>> &cxxSourceFiles, std::map<std::string, std::list<std::string>> &asmSourceFiles, std::string &linkerScript, bool checkFileExistence);
+  bool EvalFile(RteItem* file, bool checkFileExistence, const std::string& group, const std::string& base, std::string& filepath);
+  bool EvalItem(RteItem* item, bool checkFileExistence, const std::string& groupName = std::string(), const std::string& groupLayer = std::string());
   bool EvalItemTranslationControls(const RteItem* item, TranslationControlsKind kind, const std::string& groupName = std::string());
   bool GenerateRteHeaders();
   bool EvalDeviceName();
-  bool EvalFlags();
+  bool EvalFlags(bool checkFileExistence);
```
brondani, line 523: To avoid the propagation of the `checkFileExistence` argument over all these functions, would it be possible to have a `CbuildModel` class member variable instead?

tools/buildmgr/cbuildgen/src/Console.cpp
```diff
@@ -155,6 +156,7 @@ int main(int argc, const char *argv[])
     else if  (arg.compare("--quiet")                 == 0)            quiet = true;
     else if  (arg.compare(0, 12, "--pack_root=")     == 0)            packRoot = arg.substr(12, arg.length());
     else if  (arg.compare(0, 16, "--compiler_root=") == 0)            compilerRoot = arg.substr(16, arg.length());
+    else if  (arg.compare("--no-update-rte")         == 0)            updateRteFiles = false;
```
brondani, line 159: We agreed to have an inverted logic for `cbuildgen`, as commented [here](https://github.com/Open-CMSIS-Pack/devtools/pull/432#issuecomment-1273476662):
> in other words by default don't create/update RTE files but introduce an `--update-rte` flag

Please rework this behaviour, or if you prefer to provide it in a separated PR please remove the "--no-update-rte" option here.

tools/projmgr/src/ProjMgr.cpp
```diff
@@ -99,12 +100,13 @@ int ProjMgr::RunProjMgr(int argc, char **argv) {
   cxxopts::Option load("l,load", "Set policy for packs loading [latest|all|required]", cxxopts::value<string>());
   cxxopts::Option missing( "m,missing", "List only required packs that are missing in the pack repository", cxxopts::value<bool>()->default_value("false"));
   cxxopts::Option schemaCheck( "n,no-check-schema", "Skip schema check", cxxopts::value<bool>()->default_value("false"));
+  cxxopts::Option noUpdateRte( "no-update-rte", "Skip creation of RTE directory and files", cxxopts::value<bool>()->default_value("false"));
```
brondani, line 103: Since all flags have short and long options, for consistency should be accepted a short option here as well?

tools/projmgr/src/ProjMgrWorker.cpp
```diff
@@ -1876,7 +1877,7 @@ bool ProjMgrWorker::ListDependencies(vector<string>& dependencies, const string&
   set<string>dependenciesSet;
   for (const auto& selectedContext : m_selectedContexts) {
     ContextItem& context = m_contexts[selectedContext];
-    if (!ProcessContext(context, true, false)) {
+    if (!ProcessContext(context, true, false, false)) {
```
brondani, line 1880: Shouldn't the `list dependencies` command also accept and handle the `--no-update-rte` option instead of hard-coding it to `false`?

brondani (commented):
@ErikLundinSTM Thanks for this commit!
Please find my comments and suggestions near to the code.

brondani:
@ErikLundinSTM
In order to conclude this PR I am proposing to fix the integration tests:
https://github.com/ErikLundinSTM/devtools/pull/1

It seems we didn't realize we need to extend the [cbuild](https://github.com/Open-CMSIS-Pack/cbuild) options as well. Let' tackle that afterwards.

ErikLundinSTM:
> @ErikLundinSTM In order to conclude this PR I am proposing to fix the integration tests: [ErikLundinSTM#1](https://github.com/ErikLundinSTM/devtools/pull/1)

I'm struggling with the integration tests right now. Seems like it works if reverting the default behavior of `cbuildgen cmake` (`--no-update-rte` vs `--update-rte`), so I've been trying to pinpoint where to fix this in the tests, but haven't yet succeeded to find the right places.

brondani:
@ErikLundinSTM If you merge my proposed changes into your branch https://github.com/ErikLundinSTM/devtools/pull/1 all remaining integration tests failings should be solved. I have tested it in [my fork](https://github.com/brondani/devtools/actions/runs/3488161764).
If you have any question please feel free to setup a call. Thanks!

ErikLundinSTM:
> @ErikLundinSTM If you merge my proposed changes into your branch [ErikLundinSTM#1](https://github.com/ErikLundinSTM/devtools/pull/1) all remaining integration tests failings should be solved. I have tested it in [my fork](https://github.com/brondani/devtools/actions/runs/3488161764). If you have any question please feel free to setup a call. Thanks!

Sorry, I missed that PR. Is it OK if I first remove my last, temporary commit and rebase on main? Then I'll merge your fixes to the integration tests.

brondani:
> Sorry, I missed that PR. Is it OK if I first remove my last, temporary commit and rebase on main? Then I'll merge your fixes to the integration tests.

There is no need to remove the commit - that line was actually reworked in my changes. In the end we do a "squash and merge" so all the PR changes are merged into main as one single commit. And yes you can rebase on main.

brondani (approved):

edriouk (approved):

acabarbaye:
> It seems we didn't realize we need to extend the [cbuild](https://github.com/Open-CMSIS-Pack/cbuild) options as well. Let' tackle that afterwards.

Could `cbuild` be updated to have these flags available as well so that the build service can make use of this

brondani:
> Could `cbuild` be updated to have these flags available as well so that the build service can make use of this?

Yes it is tracked here: https://github.com/Open-CMSIS-Pack/cbuild/issues/21

PR #435, opened 2022-08-22
Bugfix to check if component is device dependent (#280)

grasci-arm:
follow referenced condition instead of returning false

Co-authored-by: Evgueni Driouk <evgueni.driouk@arm.com>


jkrech:
@edriouk  As far as I can see this was a regression, however it was not discovered by any of the tests. Can we please add a test now?

brondani (approved):

PR #436, opened 2022-08-23
Don't require file to exist when normalizing path

ErikLundinSTM:
Split path normalizing and existence check as a preparation to make existence check optional. This shouldn't affect behavior.

Prerequisite for #432

Signed-off-by: Erik LUNDIN <erik.lundin@st.com>

### DIFF
--- a/libs/rtefsutils/include/RteFsUtils.h
+++ b/libs/rtefsutils/include/RteFsUtils.h
@@ -248,9 +248,8 @@ public:
    * @brief normalize path by ensuring that it's absolute and canonical
    * @param path folder to be normalized
    * @param base root folder to form absolute
-   * @return true if path is normalized
   */
-  static bool NormalizePath(std::string& path, std::string base = "");
+  static void NormalizePath(std::string& path, const std::string& base = RteUtils::EMPTY_STRING);
   /**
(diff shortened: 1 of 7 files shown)


libs/rtefsutils/include/RteFsUtils.h
```diff
@@ -248,9 +248,8 @@ class RteFsUtils
    * @brief normalize path by ensuring that it's absolute and canonical
    * @param path folder to be normalized
    * @param base root folder to form absolute
-   * @return true if path is normalized
   */
-  static bool NormalizePath(std::string& path, std::string base = "");
+  static void NormalizePath(std::string& path, std::string base = "");
```
edriouk, line 252: since the signature is changed, could we also change 
std::string base = "" 
to
const std::string& = RteUtils::EMPTY_STRING

edriouk (changes requested):
Looks good, just one cosmetic remark

libs/rtefsutils/include/RteFsUtils.h
ErikLundinSTM: Fixed.

edriouk (approved):

Issue #433, opened 2022-08-18
Add generatordir to output-dirs


slhultgren:
The user/project should be able to control where generators are placing/creating files.

Proposal is to add a `generatordir` to the [output-dirs](https://github.com/Open-CMSIS-Pack/devtools/blob/main/tools/projmgr/docs/Manual/YML-Format.md#output-dirs) field.
The `generatordir` would then update the `destination` for the [generator input file](https://github.com/Open-CMSIS-Pack/devtools/blob/main/tools/projmgr/docs/Manual/YML-Format.md#workflow)

In the PDSC we can also introduce a `$G` parameter that would expand to the generator input file path so that generators easily can find it.

The Generator could still potentially create files anywhere it wants, but assuming the generator is a good citizen it can at least use the `destination` to try follow user wishes.

ReinhardKeil:
Is the proposed $G the path to the [cbuild.yml](https://github.com/Open-CMSIS-Pack/devtools/blob/main/tools/projmgr/docs/Manual/YML-CBuild-Format.md) file?

slhultgren:
> Is the proposed $G the path to the [cbuild.yml](https://github.com/Open-CMSIS-Pack/devtools/blob/main/tools/projmgr/docs/Manual/YML-CBuild-Format.md) file?

Yes, it probably should be. For now the https://github.com/Open-CMSIS-Pack/devtools/pull/462 implements $G expansion to the currently implemented `...generate.yml` file, to connect the $G to something.

Once we have the .cbuild.yml implemented, I think we can start to change $G to point to this and remove/replace the current ".generate.yml"-code.

Then we can see if we need to create a temporary more explicit copy of .cbuild.yml or not (basically just with all paths resolved absolutely everywhere instead of relying on relative/variables) but this can be discussed later, since the format would not require change, only the values.

slhultgren:
Closing this issue since it has now been implemented and merged.

PR #462, opened 2022-09-06
Add support for output-dirs[gendir] and $G

slhultgren:
Add support for `$G` variable on `<generator>` element in .pdsc format as implemented in https://github.com/Open-CMSIS-Pack/Open-CMSIS-Pack-Spec/pull/151

Add support for `gendir` on the `output-dirs` property to allow customizing the `destination` property in the generator input file as described in https://github.com/Open-CMSIS-Pack/devtools/issues/433

Contributed by STMicroelectronics.

### DIFF
--- a/libs/rtemodel/include/RteTarget.h
+++ b/libs/rtemodel/include/RteTarget.h
@@ -854,15 +854,16 @@ public:
   RteDeviceProperty* GetDeviceEnvironment() const { return m_deviceEnvironment; }
 
   /**
-   * @brief get the generator input file path
-   * @return generator input file
+   * @brief get the absolute path to the generator input file
+   * @return absolute path to the generator input file
   */
-  const std::string& GetGeneratorInputFilePath() const { return m_generatorInputFilePath; }
+  const std::string& GetGeneratorInputFile() const { return m_generatorInputFile; }
 
   /**
    * @brief set the generator input file path
+   * @param newGeneratorInputFile the new absolute path to the generator input file
   */
-  void SetGeneratorInputFilePath(const std::string& newGeneratorInputFilePath) { m_generatorInputFilePath = newGeneratorInputFilePath; }
+  void SetGeneratorInputFile(const std::string& newGeneratorInputFile) { m_generatorInputFile = newGeneratorInputFile; }
 
   /**
    * @brief get <environment> property of device with given name
@@ -961,7 +962,7 @@ protected:
   std::set<std::string> m_objects;
   std::set<std::string> m_docs;
   std::map<std::string, RteComponent*> m_scvdFiles; // component viewer description files
-  std::string m_generatorInputFilePath; // Absolute path to the generator input file
+  std::string m_generatorInputFile; // Absolute path to the generator input file
 
   // header file content
   std::set<std::string> m_RTE_Component_h; // defines put into the file
--- a/libs/rtemodel/include/RteTarget.h
+++ b/libs/rtemodel/include/RteTarget.h
@@ -854,6 +854,17 @@ public:
   RteDeviceProperty* GetDeviceEnvironment() const { return m_deviceEnvironment; }
 
   /**
+   * @brief get the generator input file path
+   * @return generator input file
+  */
+  const std::string& GetGeneratorInputFilePath() const { return m_generatorInputFilePath; }
+
+  /**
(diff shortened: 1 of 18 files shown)


libs/rtemodel/include/RteTarget.h
```diff
@@ -853,6 +853,17 @@ class RteTarget : public RteItem
   */
   RteDeviceProperty* GetDeviceEnvironment() const { return m_deviceEnvironment; }
 
+  /**
+   * @brief get the generator input file path
+   * @return generator input file
```
edriouk, line 858: the comment says the method returns file, but the method name is  GetGeneratorInputFilePath. 
It must be clear if returned string is a file or a directory,  is it absolute or relative.

libs/rtemodel/include/RteTarget.h
```diff
@@ -853,6 +853,17 @@ class RteTarget : public RteItem
...
+  */
+  const std::string& GetGeneratorInputFilePath() const { return m_generatorInputFilePath; }
```
edriouk, line 860: See previous comment: rename method to GetGeneratorInputFile() if it returns filename.

libs/rtemodel/include/RteTarget.h
```diff
@@ -853,6 +853,17 @@ class RteTarget : public RteItem
...
+
+  /**
+   * @brief set the generator input file path
+  */
```
edriouk, line 864: comment 
@param newGeneratorInputFilePath is missing

libs/rtemodel/include/RteTarget.h
```diff
@@ -853,6 +853,17 @@ class RteTarget : public RteItem
...
+  void SetGeneratorInputFilePath(const std::string& newGeneratorInputFilePath) { m_generatorInputFilePath = newGeneratorInputFilePath; }
```
edriouk, line 865: naming: see comment to GetGeneratorInputFilePath()

libs/rtemodel/include/RteTarget.h
```diff
@@ -950,6 +961,8 @@ class RteTarget : public RteItem
   std::set<std::string> m_objects;
   std::set<std::string> m_docs;
   std::map<std::string, RteComponent*> m_scvdFiles; // component viewer description files
+  std::string m_generatorInputFilePath; // Absolute path to the generator input file
```
edriouk, line 964: naming: see comment to GetGeneratorInputFilePath()

tools/projmgr/src/ProjMgrWorker.cpp
```diff
@@ -1962,21 +1967,45 @@ bool ProjMgrWorker::ExecuteGenerator(std::string& generatorId) {
...
     }
-    const string generatorWorkingDir = generator->GetExpandedWorkingDir(context.rteActiveTarget);
 
-    // Create generate.yml file with context info and destination
-    ProjMgrYamlEmitter::EmitContextInfo(context, generatorWorkingDir);
 
     error_code ec;
     const auto& workingDir = fs::current_path(ec);
-    fs::current_path(generatorWorkingDir, ec);
+    fs::current_path(generatorDestination, ec);
```
edriouk, line 2008: fs::current_path(generatorWorkingDir, ec); role was to set working directory for a generator. 
Now generator starts in the the destination directory. Is that expected for all generators?

edriouk (changes requested):
Generally looks good, however:
- clarification is needed if we deal with a file or a directory 
- tests are missing

tools/projmgr/src/ProjMgrWorker.cpp
slhultgren: Yes this was intentional on my side at least.

Before my change we used the generatorWorkingDir as current_path, and the `destination` property in the generator input file was aligned with the current_path. So the idea here is the same, since the destination is now a bit more flexible thanks to the `output-dirs[gendir]` user-definable property, the current_path should still follow `destination` which follows `gendir`.

The generator input file is currently listing the project and solution as well, so generators can still find relevant information. 

(We need to make sure this project file connection is kept when moving to the `cbuild.yml`-based format)

slhultgren:
Hi @edriouk thanks for the review and comments, all should have been addressed now, please have a look.

edriouk (approved):

PR #471, opened 2022-09-12
GitHub: Don't trigger source code checks

JonatanAntoni:
... when only documentation is touched

### DIFF
--- a/.github/workflows/buildmgr.yml
+++ b/.github/workflows/buildmgr.yml
@@ -14,6 +14,7 @@ on:
       - 'libs/rtemodel/**'
       - 'libs/rtefsutils/**'
       - 'tools/buildmgr/**'
+      - '!**/*.md'
   push:
     branches:
       - main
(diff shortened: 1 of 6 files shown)


.github/workflows/buildmgr.yml
```diff
@@ -14,6 +14,7 @@ on:
       - 'libs/rtemodel/**'
       - 'libs/rtefsutils/**'
       - 'tools/buildmgr/**'
+      - '!**/*.md'
```
soumeh01, line 17: Though it doesn't make much difference, would it be better to separate the include and exclude paths?
```
on:
  push:
    paths-ignore:
      - **/*.md
```

JonatanAntoni: Afaik, `paths` and `paths-ignore` are mutually-exclusive.

soumeh01: Ahh I see, 

`You can not use paths and paths-ignore to filter the same event in a single workflow. If you want to both include and exclude path patterns for a single event, use the paths filter along with the ! character to indicate which paths should be excluded.`

soumeh01 (approved):
LGTM

soumeh01 (approved):

jkrech (approved):
Looks good to me.

PR #480, opened 2022-09-13
[buildmgr] support target TranslateCtrls options

tarek-bochkati:


tools/buildmgr/cbuildgen/src/CMakeListsGenerator.cpp
```diff
@@ -223,67 +226,186 @@ bool CMakeListsGenerator::GenBuildCMakeLists(void) {
     }
   }
 
-  // File specific flags
```
brondani, line 226: Maybe it would be shorter to implement the code below similarly to file/group specific defines? Please see the loops between lines 69 and 104.

brondani (commented):
Could we split this PR into:
- target level
- group/file level

tarek-bochkati:
> Could we split this PR into:
> 
> * target level
> * group/file level

I propose to split the last commit in what you propose, please see my update

tools/buildmgr/cbuildgen/src/CMakeListsGenerator.cpp
tarek-bochkati: this should be addressed into separate PR

brondani (approved):

PR #492, opened 2022-09-19
[ProjMgr/BuildMgr] Add ldcflag and ldcxxflag support (#224)

yvesll:
Signed-off-by: yvesll <toyears@hotmail.com>


tools/buildmgr/cbuildgen/src/CMakeListsGenerator.cpp
```diff
@@ -46,7 +46,15 @@ bool CMakeListsGenerator::GenBuildCMakeLists(void) {
...
-  if (!m_linkerMscGlobal.empty())    cmakelists << EOL << "set(LD_FLAGS_GLOBAL \"" << CbuildUtils::EscapeQuotes(m_linkerMscGlobal) << "\")";
+  if (!m_linkerMscGlobal.empty()) {
+    cmakelists << EOL << "set(LD_FLAGS_GLOBAL \"" << CbuildUtils::EscapeQuotes(m_linkerMscGlobal);
+    if(!m_cxxFilesList.empty()) {
+      if (!m_linkerCxxMscGlobal.empty()) cmakelists << " " << CbuildUtils::EscapeQuotes(m_linkerCxxMscGlobal);
+    } else {
+      if (!m_linkerCMscGlobal.empty()) cmakelists << " " << CbuildUtils::EscapeQuotes(m_linkerCMscGlobal);
+    }
+    cmakelists << "\")";
+  }
```
soumeh01, line 57: Please add some tests generating CMakelists.txt with link-c and link-cpp flags.

tools/projmgr/schemas/common.schema.json
```diff
@@ -151,6 +151,8 @@
         "C-CPP":        { "$ref": "#/definitions/ArrayOfStrings", "description": "List of C & CPP flags" },
         "ASM":          { "$ref": "#/definitions/ArrayOfStrings", "description": "List of ASM flags"  },
         "Link":         { "$ref": "#/definitions/ArrayOfStrings", "description": "List of Linker flags" },
```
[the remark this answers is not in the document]
yvesll: Yes, to avoid introducing back incompatible changes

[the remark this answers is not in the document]
soumeh01: Why do we need `LINK `flag in the newly created projects?

My understanding says
a.	**New projects**: Shall contain both `LINK-C` and `LINK-CPP` flags (_No existing `LINK `flags_).  To support the use case for e.g say the project is currently a C project.  In this case, the linker uses `LINK-C` flags. But in the future, if the user adds CPP files to the same project. Now linker uses `LINK-CPP` flags. 
b.	**Old projects**: Shall have ONLY existing `LINK` flags. (Need to maintain it for backward compatibility).

tools/buildmgr/cbuildgen/src/CMakeListsGenerator.cpp
soumeh01, line 57: This implementation is not correct. Please see my comment https://github.com/Open-CMSIS-Pack/devtools/pull/492#discussion_r976411127

tools/buildmgr/cbuild/src/CbuildModel.cpp
```diff
@@ -1056,6 +1058,14 @@ bool CbuildModel::EvalFlags() {
...
       }
     }
+    if (ldcflags != NULL)
+    {
+      m_targetLdCFlags = SplitArgs(ldcflags->GetAttribute("add"));
+    }
+    if (ldcxxflags != NULL)
+    {
+      m_targetLdCxxFlags = SplitArgs(ldcxxflags->GetAttribute("add"));
+    }
```
soumeh01, line 1068: Please consider `flag ` attribute in C and CPP flags

tools/projmgr/test/data/TestSolution/test.csolution.yml
```diff
@@ -9,6 +9,8 @@ solution:
           C-CPP: [-C-CPP-type-CM0]
           ASM: [-ASM-type-CM0]
           Link: [-Link-type-CM0]
+          Link-C: [-Link-C-type-CM0]
+          Link-CPP: [-Link-CPP-type-CM0]
```
soumeh01, line 13: Can you add some examples for multiple Linker flags?

tools/buildmgr/test/testinput/Examples/GCC/Build_GCC/CMakeLists.txt.ref
```diff
@@ -17,7 +17,7 @@ set(MVE NO_MVE)
 set(BYTE_ORDER Little-endian)
 set(AS_FLAGS_GLOBAL "-Wa,--gdwarf-2 -mthumb")
 set(CC_FLAGS_GLOBAL "-O -Wall -gdwarf-2 -mapcs-frame -mthumb")
-set(LD_FLAGS_GLOBAL "--entry=Reset_Handler --specs=nosys.specs -mcpu=cortex-m3 -mthumb")
+set(LD_FLAGS_GLOBAL "--entry=Reset_Handler --specs=nosys.specs -mcpu=cortex-m3 -mthumb -static")
```
soumeh01, line 20: I have a question. In general, does the order of the flags matter?
@yvesll @brondani

tools/buildmgr/test/testinput/Examples/GCC/Build_GPP/Simulation.cprj
```diff
@@ -19,6 +19,7 @@
   <target Ddsp="NO_DSP" Dendian="Little-endian" Dfpu="NO_FPU" Dmve="NO_MVE" Dname="ARMCM3" Dsecure="Non-secure" Dtz="NO_TZ" Dvendor="ARM:82">
     <output name="Build_G++" type="exe"/>
     <ldflags add="--entry=Reset_Handler --specs=nosys.specs -lstdc++ -mcpu=cortex-m3 -mthumb" compiler="GCC" file="./RTE/Device/ARMCM3/gcc_arm.ld"/>
+    <ldcxxflags add="-static" compiler="GCC"/>
```
soumeh01, line 22: We are missing tests around scenarios mentioned [here](https://github.com/Open-CMSIS-Pack/devtools/issues/224#issuecomment-1274697707)

tools/buildmgr/test/testinput/Examples/GCC/Build_GCC/CMakeLists.txt.ref
brondani: Yes it would be important to preserve the flags order. For example single-pass linkers may need the libraries in the correct order to be able to find the symbols.

yvesll: So, we must keep the order like that in csolution/cproject file. If LINK-CPP was added before LINK in , flags under LINK-CPP should also be added before LINK in LD_FLAGS_GLOBAL, is my understanding correct? @brondani

brondani: @yvesll Keeping the relative order between `Link` and `Link-CPP` as they are in `yml` files can be tricky because the `misc` field is accepted in several places. I would say it would be predictable enough when `Link` flags appear in the command line before the `Link-CPP` flags as it is in the [doc](https://github.com/Open-CMSIS-Pack/devtools/blob/main/tools/projmgr/docs/Manual/YML-Input-Format.md#misc).

tools/buildmgr/cbuild/src/CbuildModel.cpp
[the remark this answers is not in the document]
yvesll: I think the adoption of `ldcflag` or `ldcxxflag` should only depend on whether cpp source exists. And it seems this satisfies the user scenario, as the Cmakefile generator will always find the correct option.
For linker flag types, it seems only 'compiler', 'file', and 'add' attributes be used. And file attribute for the linker script will still keep in `ldflags` even if the `LINK` option is empty.

tools/projmgr/test/data/TestSolution/test.csolution.yml
yvesll: Does this make sense? Flags under misc share the same API to transfer to cprj format.

tools/buildmgr/test/testinput/Examples/GCC/Build_GPP/Simulation.cprj
yvesll: Added [here](https://github.com/Open-CMSIS-Pack/devtools/pull/492/commits/73e7e7c04061bf4a1da262e693aa4a7204e6b561#diff-79dce1734fc53449a5cc900eb03c8984707c08804d1183b9d44d1becd45cce1a)

tools/projmgr/src/ProjMgrWorker.cpp
```diff
@@ -2169,13 +2171,15 @@ bool ProjMgrWorker::ProcessSequencesRelatives(ContextItem& context, BuildType& b
...
   for (auto& misc : build.misc) {
-    if (!ProcessSequencesRelatives(context, misc.as)    ||
-        !ProcessSequencesRelatives(context, misc.as)    ||
-        !ProcessSequencesRelatives(context, misc.c)     ||
-        !ProcessSequencesRelatives(context, misc.cpp)   ||
-        !ProcessSequencesRelatives(context, misc.c_cpp) ||
-        !ProcessSequencesRelatives(context, misc.lib)   ||
-        !ProcessSequencesRelatives(context, misc.link)) {
+    if (!ProcessSequencesRelatives(context, misc.as)     ||
+        !ProcessSequencesRelatives(context, misc.as)     ||
```
soumeh01, line 2175: Can you please remove one of the `ProcessSequencesRelatives(context, misc.as)`? Seems like this is a mistake.

yvesll: Sorry, it's my fault when rebasing the main branch. THanks for the careful catch!

soumeh01 (approved):

brondani (approved):

yvesll:
Hi, @brondani, one small question, do I need to rebase my commits into one before the merge?

PR #504, opened 2022-09-30
Fix $defs.ComponentType.properties.files.items type

ithinuel:
items should be a object instead of an array.


soumeh01 (approved):
Correct. Thanks for pointing out this

brondani (approved):

PR #553, opened 2022-10-25
[buildmgr] support translate ctrls options with GCC

tarek-bochkati:


tools/buildmgr/cbuildgen/config/GCC.11.2.1.cmake
```diff
@@ -15,6 +15,77 @@ set(CC ${TOOLCHAIN_ROOT}/${PREFIX}gcc${EXT})
...
+    set(${flags} "${flag} ${${flags}}" PARENT_SCOPE)
+  else()
+    string(TOLOWER "${option}" _option)
+    message(FATAL_ERROR "unkown '${_option}' value '${value}' !")
+  endif()
+endfunction()
+
+function(cbuild_set_options_flags lang optimize debug warnings flags)
+  set(opt_flags)
+  # GCC provide a better optimization level for debug
```
tarek-bochkati, line 79: is it fine to have this tweek, or can just have `-O0`

tools/buildmgr/cbuildgen/config/GCC.11.2.1.cmake
```diff
@@ -15,6 +15,77 @@ set(CC ${TOOLCHAIN_ROOT}/${PREFIX}gcc${EXT})
...
+      string(APPEND TMP_DEFINES "-Wa,-defsym,\"${KEY}=${VALUE}\" ")
+    else()
+      string(APPEND TMP_DEFINES "-D${ENTRY} ")
+    endif()
+  endforeach()
+  set(${defines} ${TMP_DEFINES} PARENT_SCOPE)
+endfunction()
+
+set(OPTIMIZE_VALUES       "debug" "none" "balanced" "size" "speed")
+set(OPTIMIZE_CC_FLAGS     "-Og"   "-O0"  "-O2"      "-Oz" "-O3")
```
brondani, line 42: What is the rational for using `-Oz` for `size`? Maybe it would be safer to use `-Os` which is less aggressive, similarly about using `O3` and not `Ofast` for `speed`?

tools/buildmgr/cbuildgen/config/GCC.11.2.1.cmake
brondani: In my opinion it is fine but it should not overwrite an user selection. In other words I would check whether `optimize` is empty instead of `none`.

brondani (commented):
@tarek-bochkati Thanks for the PR.
Are you planning to add test cases to this PR or separately?

tools/buildmgr/cbuildgen/config/GCC.11.2.1.cmake
tarek-bochkati: I have thought I can use the best size optimization possible.
But let's shoot for less-aggressive size optimization.

tools/buildmgr/cbuildgen/config/GCC.11.2.1.cmake
tarek-bochkati: your proposal seems better

tarek-bochkati:
> Are you planning to add test cases to this PR or separately?
@brondani 
I would prefer having a separate PRs, it will be easier for me to progress with tiny PRs

brondani (approved):

PR #561, opened 2022-10-26
[buildmgr] support translate ctrls options with AC5/AC6

tarek-bochkati:


jeromecoutant:
Tested OK with my projects with AC6 and:
  optimize: size
  debug: on

jeromecoutant:
But failed with
- debug: off
- warnings: off
:-(

tools/buildmgr/cbuildgen/config/AC6.6.18.0.cmake
```diff
@@ -46,6 +46,53 @@ function(cbuild_set_defines lang defines)
   set(${defines} ${TMP_DEFINES} PARENT_SCOPE)
 endfunction()
 
+set(OPTIMIZE_VALUES       "none" "balanced" "size" "speed")
+set(OPTIMIZE_CC_FLAGS     "-O0"  "-O2"      "-Os" "-Ofast")
```
brondani, line 50: According to the [Arm Compiler Reference Manual](https://developer.arm.com/documentation/101754/0619/armclang-Reference/armclang-Command-line-Options/-O--armclang-?lang=en):
`s` _Performs optimizations to reduce code size, balancing code size against code speed._
`z` _Performs optimizations to minimize image size._
`Omin`, `Omax` and `Ofast` are recommended with caution or might violate strict compliance, so I would avoid them.

For this reason I would suggest the following, would you agree?
```suggestion
set(OPTIMIZE_CC_FLAGS     "-O0"  "-Os"      "-Oz" "-O3")
```

tarek-bochkati:
> But failed with
> 
> * debug: off
> * warnings: off
>   :-(

@jeromecoutant thanks a lot for this test, this revealed that copying a list to another looses the empty values, I will commit a proper fix ASAP

jeromecoutant (dismissed):
Verified with IAR, AC6 and GCC
with
+  optimize: balanced
+  debug: on
+  warnings: on

and
+  optimize: size
+  debug: off
+  warnings: off

tools/buildmgr/cbuildgen/config/AC6.6.18.0.cmake
tarek-bochkati: @brondani thanks for you feedback.

according to ARM-CLANG 6.18 documentation:
`-Omin` is the most aggressive size optimization, which enable **LTO**
also with GCC we can enable LTO optimization using gold linker plugins

I would like to avoid exotic optimizations (IAR: no-size constraints, vanilla or proprietary LTO) into the standard,
since I believe the user have to specify them directly into flags (because this does not guarantee to be fully standards-compliant for all code cases)

so to be fair with GCC, I recommend to use similar option `-Oz`, which you suggested for AC6

brondani (approved):

brondani:
@tarek-bochkati Thanks for this PR, please proceed with the test cases.

Issue #469, opened 2022-09-08
PACKCHK: wrong M371 error?


jeromecoutant:
Hi

```
*** ERROR M371: C:/xxx.pdsc (Line 761)
  Device Feature for 'STM32U599/5A9': 'MIPI' unknown.
```
We think it should be correct as MIPI is present in pack.xsd:
https://github.com/Open-CMSIS-Pack/Open-CMSIS-Pack-Spec/blob/main/schema/PACK.xsd#L896

Regards,

@tarek-bochkati

jkrech:
This is a `packchk` defect. The features listed in the Pack Schema is most up-to-date and `packchk` needs to catch up. 
We should review whether packchk should avoid duplicating checks which are covered by the schema check.

jeromecoutant:
packchk could include schema check ?

jeromecoutant:
Correction is linked to #506? Is this correct ?

thorstendb-ARM:
MIPI is part of PackChk since 26ad5ecce580c2e39c9e0b1548cba3e28386e83d :
`  m_featureTableDevice["MIPI"] = FeatureEntry("MIPI", "<feature type=\"MIPI\"/>", "MIPI");`

jeromecoutant:
> MIPI is part of PackChk since https://github.com/Open-CMSIS-Pack/devtools/commit/26ad5ecce580c2e39c9e0b1548cba3e28386e83d

Which pachchk version ?

thorstendb-ARM:
> > MIPI is part of PackChk since [26ad5ec](https://github.com/Open-CMSIS-Pack/devtools/commit/26ad5ecce580c2e39c9e0b1548cba3e28386e83d)
> 
> Which pachchk version ?

Not yet released, will do after https://github.com/Open-CMSIS-Pack/devtools/issues/657
https://github.com/Open-CMSIS-Pack/devtools/releases

thorstendb-ARM:
see PackChk 1.3.98
https://github.com/Open-CMSIS-Pack/devtools/releases/tag/tools%2Fpackchk%2F1.3.98

jeromecoutant:
Verified

PR #568, opened 2022-11-02
GitHub#469 (#362)

grasci-arm:
added missing features from Device and Board table:
https://github.com/Open-CMSIS-Pack/Open-CMSIS-Pack-Spec/blob/main/schema/PACK.xsd#L932

### DIFF
--- a/tools/packchk/src/ValidateSyntax.cpp
+++ b/tools/packchk/src/ValidateSyntax.cpp
@@ -345,7 +345,7 @@ bool ValidateSyntax::CheckPackageReleaseDate(RtePackage* pKg)
   string latestVersion;
   string latestDate;
   int latestLineNo = 0;
-  
+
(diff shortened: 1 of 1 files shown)


JonatanAntoni (approved):

jeromecoutant:
Hi
Tip: rather than indicating issues numbers in PR header, you should add in the description `fix #469`
Then GitHub is able to add automatically some link between tickets, and close the issue when PR is merged

JonatanAntoni:
@jeromecoutant, thanks for pointing this out. I am not sure about using "fix" because to my knowledge this would auto-close the referenced issue. Do we want that? I'd rather expect the owner of an issue to close it once the solution is satisfying.

In PRs we need to be careful not to use short #-numbers but prefix with `<org>/<repo>/#<nr>` to make the references work across forks.

PR #575, opened 2022-11-04
[buildmgr] add translate ctrls options test using GCC

tarek-bochkati:


brondani (commented):
@tarek-bochkati Thanks for this PR.
Would it be possible to add or extend test cases to cover the toolchain mapping/dictionary that was introduced?
For example by using the function [CBuildTestFixture::CheckCompileCommand](https://github.com/Open-CMSIS-Pack/devtools/blob/main/tools/buildmgr/test/integrationtests/src/CBuildTestFixture.cpp#L148)
Thanks!

tarek-bochkati:
> @tarek-bochkati Thanks for this PR. Would it be possible to add or extend test cases to cover the toolchain mapping/dictionary that was introduced? For example by using the function [CBuildTestFixture::CheckCompileCommand](https://github.com/Open-CMSIS-Pack/devtools/blob/main/tools/buildmgr/test/integrationtests/src/CBuildTestFixture.cpp#L148) Thanks!

Thanks for pointing out the function, I was checking for it, but no luck.
I'm on it.

brondani (approved):

PR #583, opened 2022-11-10
Stop Obsolete Workflows

grasci-arm:


soumeh01 (commented):
We can save computation time by stopping the obsolete workflows in the scenario when the user has created a PR and just after that he/she realizes that there are some more changes that need to go. By default, the old workflow keeps on running, while your last push initiated another one which is a waste of resources.
With this change, once a new child spawned we can stop the old running redundant workflows.

https://docs.github.com/en/actions/using-jobs/using-concurrency

JonatanAntoni (approved):

PR #661, opened 2023-01-13
Add issue templates

JonatanAntoni:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,26 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: 'bug'
```
jkrech, line 5: Can we add a "severity" or "impact" level  to choose from?

JonatanAntoni: GitHub issues don't have such a thing, it only has "labels". We could add additional labels for "severity" but one would need to deliberately add one of these. The template only specifies the presets but not any kind of "please choose one out of ...".

jkrech: It is understood that we need to map everything to labels :(

There is "issue forms" https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms
The feature is in "beta". It looks like this could come handy for mandating information as well as selecting options from a drop down.

jkrech (approved):
We should consider revising introducing severity levels for bugs as labels.

Issue #434, opened 2022-08-19
*.cbuild-index.yml + *.cbuild.yml as long term replacement for *.CPRJ + Proposal for use as lockfile + Generator input


ReinhardKeil:
This proposal is to replace (over time) the `*.CPRJ` file(s) with a single `*.cbuild.yml` file that:

 - contains the build instructions for the whole application.  A new `cbuild` tool could selectively generate one project or all projects.
 - uses controls that are identical with csolution.yml, but no `build/target types` (and the related `for-type:`)
 - compiler is fixed for a project (therefore no `for-compiler:`)
 - component files are explicitly stated (which removes the need to manage components at `cbuild` tool-level)
 - additional information (such as `from-pack:`) is stated for each component
 - `clayer.yml`/`cdefault.yml` information is included (no reference to other files - all self-contained in `*.cbuild.yml` file)
 - lists all `packs:` with explicit version number (this can be used as loop-back file to csolution tool to lock the pack versions)

The file would be generated by the `csolution` tool.

Example: 
```yml
# section that defines global setup for all projects (initally only packs: would be listed here)

  packs:  # all packs used in the projects (exception when a project uses a different pack version the packs: would be also under project)
    - pack: ARM::CMSIS@5.9.0
    - pack: ARM::CMSIS-Driver@2.7.2
    - pack: ARM::CMSIS-FreeRTOS@10.4.6
    - pack: ARM::mbedTLS@1.7.0
    - pack: AWS::AWS_IoT_Over-the-air_Update@3.0.0-Beta
    - pack: AWS::backoffAlgorithm@1.0.0-Beta
    - pack: AWS::coreJSON@3.0.0-Beta
    - pack: AWS::coreMQTT@1.1.0-Beta
    - pack: AWS::coreMQTT_Agent@1.0.1-Beta
    - pack: AWS::corePKCS11@3.0.0-Beta
    - pack: AWS::FreeRTOS-Plus-TCP@2.3.2-Beta
    - pack: Arm-Packs::PKCS11@1.0.0
    - pack: Keil::ARM_Compiler@1.3.0
    - pack: MDK-Packs::IoT_Socket@1.3.0
    - pack: MDK-Packs::TinyCBOR@1.0.1
    - pack: ARM::V2M_MPS3_SSE_300_BSP@1.2.0    # DFP and BSP packs might need some special handling
    - pack: Keil::STM32U5xx_DFP@2.0.0
[...]

PR #673, opened 2023-01-19
cbuild.yml as generator input

ErikMallbergSTM:
Implements required things in the cbuild.yml according to [Generator (Proposal).md](https://github.com/Open-CMSIS-Pack/devtools/blob/main/tools/projmgr/docs/Manual/Generator%20(Proposal).md) and https://github.com/Open-CMSIS-Pack/devtools/issues/434#issuecomment-1337608400

Adds the following information when generating `*.cbuild.yml` files:
- csolution
- cproject
- input to component generators:
  * generator command
  * arguments
  * files
  * gpdsc file
  * working directory

In addition, this switches to use `*.cbuild.yml` as input to generators (`$G` set correspondingly)

Limitations/future work:
- In addition to device, board should probably be added to the build description schema and implementation
- GPDSC is not used to update `*.cbuild.yml`
- Working dir is inconsistent between `ProjMgrWorker::ExecuteGenerator()` and what is recorded in `*.cbuild.yml` (`gendir` should not be used for working dir, only destination)

Signed-off-by: Erik MÅLLBERG <erik.lundin@st.com>

### DIFF
--- a/libs/rtemodel/include/RteGenerator.h
+++ b/libs/rtemodel/include/RteGenerator.h
@@ -44,6 +44,12 @@ public:
   virtual ~RteGenerator() override;
 
   /**
+   * @brief get generator commands for all host types without expansion
+   * @return map with host type ("win", "linux", "mac", "other", or "all") as key and generator command as value
+  */
+  std::map<std::string, std::string> GetCommands() const;
+
+  /**
    * @brief get generator command without expansion
    * @param hostType host type to match, empty to match current host
    * @return generator command for specified host type
@@ -127,6 +133,20 @@ public:
  * @return vector of arguments consisting of switch and value in pairs
 */
   std::vector<std::pair<std::string, std::string> > GetExpandedArguments(RteTarget* target, const std::string& hostType = EMPTY_STRING) const;
+
+  /**
+   * @brief get the absolute path to the executable with expanded key sequences for specified target
+   * @param target pointer to RteTarget
+   * @return expanded executable
+  */
+  std::string GetExecutable(RteTarget* target) const;
+
+  /**
+   * @brief get the absolute paths to executables for all host types with the expanded key sequences for specified target
+   * @param target pointer to RteTarget
+   * @return map with host type ("win", "linux", "mac", "other" or "all") as key and path to executable as value
+  */
+  std::map<std::string, std::string> GetExecutables(RteTarget* target) const;
 
   /**
    * @brief get full command line with arguments and expanded key sequences for specified target
--- a/libs/rtemodel/include/RteGenerator.h
+++ b/libs/rtemodel/include/RteGenerator.h
@@ -121,12 +121,12 @@ public:
   const std::string& GetWorkingDir() const { return GetItemValue("workingDir"); }
 
  /**
- * @brief get all arguments as vector similar to argv for the given host type
+ * @brief get all arguments as vector for the given host type
  * @param target pointer to RteTarget
  * @param hostType host type, empty to match current host
- * @return vector of arguments including command line at 0 vector element
+ * @return vector of arguments consisting of switch and value in pairs
 */
-  std::vector<std::string> GetExpandedArgv(RteTarget* target, const std::string& hostType = EMPTY_STRING) const;
+  std::vector<std::pair<std::string, std::string> > GetExpandedArguments(RteTarget* target, const std::string& hostType = EMPTY_STRING) const;
 
   /**
    * @brief get full command line with arguments and expanded key sequences for specified target
--- a/libs/rtemodel/include/RteGenerator.h
+++ b/libs/rtemodel/include/RteGenerator.h
@@ -45,9 +45,18 @@ public:
 
   /**
    * @brief get generator command without expansion
-   * @return generator command
+   * @param hostType host type to match, empty to match current host
+   * @return generator command for specified host type
   */
-  const std::string GetCommand() const;
+  const std::string GetCommand(const std::string& hostType = EMPTY_STRING ) const;
+
+  /**
+  * @brief get expanded generator executable command
+  * @param target pointer to RteTarget
+  * @param hostType host type to match, empty to match current host
+  * @return generator command for specified host type
+ */
+  std::string GetExecutable(RteTarget* target, const std::string& hostType = EMPTY_STRING) const;
 
   /**
    * @brief get item containing command line arguments
@@ -105,18 +114,27 @@ public:
   */
   const std::string& GetGpdsc() const;
 
-  /**
-   * @brief get generator working directory
-   * @return working directory value
-  */
+ /**
+  * @brief get generator working directory
+  * @return working directory value
+ */
   const std::string& GetWorkingDir() const { return GetItemValue("workingDir"); }
+
+ /**
+ * @brief get all arguments as vector similar to argv for the given host type
+ * @param target pointer to RteTarget
+ * @param hostType host type, empty to match current host
+ * @return vector of arguments including command line at 0 vector element
+*/
+  std::vector<std::string> GetExpandedArgv(RteTarget* target, const std::string& hostType = EMPTY_STRING) const;
 
   /**
    * @brief get full command line with arguments and expanded key sequences for specified target
    * @param target pointer to RteTarget
+   * @param hostType host type, empty to match current host
    * @return expanded command line with arguments, properly quoted
   */
-  std::string GetExpandedCommandLine(RteTarget* target) const;
+  std::string GetExpandedCommandLine(RteTarget* target, const std::string& hostType = EMPTY_STRING) const;
 
   /**
    * @brief get absolute path to gpdsc file for specified target
--- a/libs/rtemodel/src/RteGenerator.cpp
+++ b/libs/rtemodel/src/RteGenerator.cpp
@@ -88,14 +88,14 @@ RteItem* RteGenerator::GetArgumentsItem(const string& type) const
   return args;
 }
 
-const string RteGenerator::GetCommand() const {
+const string RteGenerator::GetCommand(const std::string& hostType) const {
 
   RteItem* exe = GetItemByTag("exe");
   if (exe != NULL) {
     const list<RteItem*>& children = exe->GetChildren();
     for (auto it = children.begin(); it != children.end(); it++) {
       RteItem* cmd = *it;
-      if (cmd->GetTag() != "command" || !cmd->MatchesHost()) {
+      if (cmd->GetTag() != "command" || !cmd->MatchesHost(hostType)) {
         continue;
       }
       const string& key = cmd->GetAttribute("key");
@@ -120,11 +120,9 @@ const string RteGenerator::GetCommand() const {
   return GetItemValue("command");
 }
 
-
-string RteGenerator::GetExpandedCommandLine(RteTarget* target) const
+string RteGenerator::GetExecutable(RteTarget* target, const std::string& hostType) const
 {
-  string fullCmd;
-  string cmd = GetCommand();
+  string cmd = GetCommand(hostType);
   if (cmd.empty())
     return cmd;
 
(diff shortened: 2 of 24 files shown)


ErikMallbergSTM:
@brondani Please have a look if this matches what we discussed.

tools/projmgr/src/ProjMgrYamlEmitter.cpp
```diff
@@ -229,6 +235,67 @@ void ProjMgrYamlCbuild::SetComponentFilesNode(YAML::Node node, const ContextItem
...
+        execIt = executables.find("all");
+      if (execIt == executables.end())
+        continue;
+      commandNode[YAML_FILE] = FormatPath(execIt->second, context->directories.cprj);
+
+      // Arguments
+      YAML::Node argumentsNode;
+      for (const auto& argument : arguments) {
+        if (argument.host == host || argument.host == "all" || argument.host == "") {
+          string arg = argument.argument;
```
brondani, line 277: Maybe the `argument` element of the `GeneratorArgument struct` could be renamed to avoid such pattern `argument.argument`, for example `argument.body`?

brondani (commented):
@ErikMallbergSTM Many thanks for this contribution!
Please take a look at my comments.

@edriouk Could you please review the changes under `rtemodel`?

libs/rtemodel/include/RteGenerator.h
```diff
@@ -25,6 +25,18 @@ class RteTarget;
...
 
+/**
+ * @brief generator argument containing
+ *        the actual argument,
+ *        the argument switch if any (empty otherwise),
+ *        a host type if any ("win", "linux", "mac", "other" or "all")
+*/
+struct GeneratorArgument {
+  std::string argument;
+  std::string switch_;
```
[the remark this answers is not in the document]
edriouk: I do not see any need in this structure. The data is already available in RteGenerator object and can be accessed by 
RteGenerator::GetExpandedCommandLine() and RteGenerator::GetArgumentsItem() calls

libs/rtemodel/src/RteGenerator.cpp
```diff
@@ -139,32 +200,33 @@ string RteGenerator::GetExpandedCommandLine(RteTarget* target) const
     if (p)
       cmd = p->GetAbsolutePackagePath() + cmd;
   }
+  return cmd;
+}
 
-  RteItem* argsItem = GetArgumentsItem("exe");
-  if (argsItem) {
+string RteGenerator::GetExpandedCommandLine(RteTarget* target) const
```
edriouk, line 206: Is there any change in the logic apart from using an intermediate structure?

edriouk (changes requested):

tools/projmgr/schemas/common.schema.json
```diff
@@ -12,14 +12,17 @@
     },
     "BuildContext": {
       "type": "string",
-      "pattern": "^(((^[.][^.+\\s]*)|(^[+][^.+\\s]*))|((^[.][^.+\\s]*)[+][^.+\\s]*|(^[+][^.+\\s]*)[.][^.+\\s]*))$"
+      "pattern": "^[^.+\\s]*(\\.[^.+\\s]+)?(\\+[^.+\\s]+)?$"
```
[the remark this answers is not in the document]
ErikMallbergSTM: The description in the JSON schema (`[.<build-type>][+<target-type>]`) in combination with the implementation in `ProjMgrWorker::AddContext` led me to think that build type always comes before target type. I have now changed the pattern to accept build type and target type in arbitrary order.

tools/projmgr/schemas/common.schema.json
```diff
@@ -105,6 +108,11 @@
       "minItems": 1,
       "items": {"type": "object", "additionalProperties": {"type": "string"}}
     },
+    "VersionType": {
+      "type": "string",
+      "pattern": "^\\d+\\.\\d+\\.\\d+(-[0-9A-Za-z-]+(\\.[0-9A-Za-z-]+)*)?(\\+[0-9A-Za-z-]+(\\.[0-9A-Za-z-]+)*)?$",
```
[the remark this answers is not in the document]
ErikMallbergSTM: According to [the spec](https://open-cmsis-pack.github.io/Open-CMSIS-Pack-Spec/main/html/pdsc_package_pg.html#VersionType), only alphanumeric characters and hyphen are allowed in the version type. `\w` also includes underscore (`[A-Za-z0-9_]`).

libs/rtemodel/include/RteGenerator.h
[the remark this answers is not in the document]
ErikMallbergSTM: @brondani The reason of the underscore was that `switch` is a reserved word in C++. I'm open for other names (but if we skip this struct I guess it doesn't matter).

@edriouk `RteGenerator::GetExpandedCommandLine()` returns a string with the full command line (executable matching the current host plus arguments), while I wanted something more structured. To me, it's natural to utilize the structure of YAML to keep the executable and the individual arguments separated (which in turn allows a consumer to treat arguments as separate strings without going down the rabbit hole of proper quoting). Regarding `RteGenerator::GetArgumentsItem()`,  I'm open to use that instead if you prefer that we don't add `RteGenerator::GetArguments()`. It just seemed to make sense to have a function to return arguments in a "refined" form.

tools/projmgr/src/ProjMgrYamlEmitter.cpp
```diff
@@ -229,6 +235,67 @@ void ProjMgrYamlCbuild::SetComponentFilesNode(YAML::Node node, const ContextItem
...
+  for (const auto& [generatorId, generator] : context->generators) {
+    YAML::Node genNode;
+    const string& workingDir = generator->GetExpandedWorkingDir(context->rteActiveTarget);
+    SetNodeValue(genNode[YAML_WORKING_DIR], FormatPath(workingDir, context->directories.cprj));
+    const string& gpdscFile = generator->GetExpandedGpdsc(context->rteActiveTarget);
+    SetNodeValue(genNode[YAML_GPDSC], FormatPath(gpdscFile, context->directories.cprj));
+
+    const auto& executables = generator->GetExecutables(context->rteActiveTarget);
+    const auto& arguments = generator->GetArguments("exe");
+    for (const string& host : {"win", "linux", "mac", "other"}) {
```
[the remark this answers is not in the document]
ErikMallbergSTM: Fixed.

tools/projmgr/src/ProjMgrWorker.cpp
```diff
@@ -2623,16 +2633,10 @@ bool ProjMgrWorker::ExecuteGenerator(std::string& generatorId) {
       generatorDestination += '/';
     }
 
-    const auto generatorInputFilePath = ProjMgrYamlEmitter::EmitContextInfo(context, generatorDestination);
```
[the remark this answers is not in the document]
ErikMallbergSTM: Oh, so that's why... Fixed by removing the `EmitContextInfo` implementation.

tools/projmgr/src/ProjMgrYamlEmitter.cpp
ErikMallbergSTM: Sure. But we may not even keep that struct in the end. See discussion with @edriouk above.

libs/rtemodel/src/RteGenerator.cpp
ErikMallbergSTM: Not really. It just seemed to make sense to have a function to return arguments in a "refined" form.

libs/rtemodel/include/RteGenerator.h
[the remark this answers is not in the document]
ErikMallbergSTM: I have now removed `RteGenerator::GetArguments()` and the struct `GeneratorArgument` and call `RteGenerator::GetArgumentsItem()` from `ProjMgrYamlCbuild::SetGeneratorsNode()`.

libs/rtemodel/src/RteGenerator.cpp
ErikMallbergSTM: I have now removed `GetArguments()` (see discussion above).

libs/crossplatform/include/CrossPlatformUtils.h
```diff
@@ -69,6 +69,13 @@ class CrossPlatformUtils
   */
   static const std::string& GetHostType();
 
+  /**
+   * @brief Checks if the given host type matches the OS this code is built for: Linux, Windows or Mac OS
+   * @param host type given as "win", "linux", "mac", "all" or "" (the last two always match)
+   * @return true if the given host type matches the host platform
+  */
+  static bool MatchesHost(const std::string& hostType);
```
edriouk, line 77: The method does not belong to cross-platform library. The constants "win", "linux", "mac", "all" are specific to CMSIS standard and should only be used in that context.

libs/rtemodel/include/RteGenerator.h
```diff
@@ -111,6 +117,20 @@ class RteGenerator : public RteItem
...
+   * @return expanded executable
+  */
+  std::string GetExecutable(RteTarget* target) const;
+
+  /**
+   * @brief get the absolute paths to executables for all host types with the expanded key sequences for specified target
+   * @param target pointer to RteTarget
+   * @return map with host type ("win", "linux", "mac", "other" or "all") as key and path to executable as value
+  */
+  std::map<std::string, std::string> GetExecutables(RteTarget* target) const;
```
edriouk, line 132: I still do not understand why all commands are needed in a separate variable? Where should this method be used?

libs/rtemodel/include/RteGenerator.h
```diff
@@ -43,6 +43,12 @@ class RteGenerator : public RteItem
   */
   virtual ~RteGenerator() override;
 
+  /**
+   * @brief get generator commands for all host types without expansion
+   * @return map with host type ("win", "linux", "mac", "other", or "all") as key and generator command as value
+  */
+  std::map<std::string, std::string> GetCommands() const;
```
edriouk, line 50: The same as for GetExecutables() : what is the purpose of this method?

edriouk (changes requested):

libs/crossplatform/include/CrossPlatformUtils.h
ErikMallbergSTM: The string taken as parameter by `CrossPlatformUtils::MatchesHost()` exactly matches what is returned by `CrossPlatformUtils::GetHostType()`. I don't see how that is different. However, since this function isn't used anymore (since I removed `RteGenerator::GetArguments()`), I could remove it.

libs/rtemodel/include/RteGenerator.h
ErikMallbergSTM: `RteGenerator::GetExecutables()` is used from `ProjMgrYamlCbuild::SetGeneratorsNode()`. I'm not sure what you suggest that we use instead, but to me it made sense to have a method providing the executable files of a generator.

libs/rtemodel/include/RteGenerator.h
ErikMallbergSTM: Rather than implementing the same functionality (taking an `RteItem` object and extracting the list of commands) in both `RteGenerator` and `ProjMgrYamlCbuild`, I thought it would make sense for a generator to return it's commands (and the now removed arguments).

libs/crossplatform/include/CrossPlatformUtils.h
ErikMallbergSTM: It kind of fulfills the same purpose as `RteGenerator::GetExpandedCommandLine()`. That function didn't fulfill my needs, so I wanted the same information but in a more usable form than a text blob.

libs/rtemodel/include/RteGenerator.h
ErikMallbergSTM: It kind of fulfills the same purpose as `RteGenerator::GetExpandedCommandLine()`. That function didn't fulfill my needs, so I wanted the same information but in a more usable form than a text blob.

libs/rtemodel/include/RteGenerator.h
edriouk: I wonder why do you need commands for all platforms, say running on Linux?

ErikMallbergSTM: Because the lockfile we generate shouldn't be different based on which platform it's generated. It should contain the commands for all platforms.

edriouk: But expanded paths are absolute, how can it work for different platforms?

ErikMallbergSTM: By using `ProjMgrYamlCbuild::FormatPath()` on the executable in `ProjMgrYamlCbuild::SetGeneratorsNode()`.

libs/crossplatform/include/CrossPlatformUtils.h
edriouk: I agree.

ErikMallbergSTM: Should be removed now.

edriouk:
@ErikMallbergSTM, please find attached my suggestion for  RteGenerator and related RteModel files
Would it fit you?
Please note that RteItem::ExpandString() calls RteCallback::ExpandString(). 
You can override that in ProjMgrCallback to call FormatPath or similar methods there.
This will avoid problems with switches and quotes.

[PR673_RteModelChanes.txt](https://github.com/Open-CMSIS-Pack/devtools/files/10557150/PR673_RteModelChanes.txt)

ErikMallbergSTM:
> @ErikMallbergSTM, please find attached my suggestion for RteGenerator and related RteModel files Would it fit you? Please note that RteItem::ExpandString() calls RteCallback::ExpandString(). You can override that in ProjMgrCallback to call FormatPath or similar methods there. This will avoid problems with switches and quotes.
> 
> [PR673_RteModelChanes.txt](https://github.com/Open-CMSIS-Pack/devtools/files/10557150/PR673_RteModelChanes.txt)

I have applied the patch and adapted the code in `ProjMgrYamlEmitter` accordingly (but not pushed yet). The application compiles, but I have found some problems:
- Quoting the arguments is counterproductive when they should be added to a YAML structure. The quotes aren't needed to keep the arguments separated and become escaped when serialized to YAML anyway, making the argument invalid
- Having the switch already prepended to the argument strings make it difficult to know if an argument is an absolute path that should be sent to `FormatPath()`

edriouk:
@ErikMallbergSTM , 
1. I have suggested putting FormatPath or similar in ExpandString() override to make it easier.
2. You must address quoting problems anyway since an argument element can contain something like:
`<argument>-I"$P/My Include Path Wish Spaces" </argument>`

ErikMallbergSTM:
@edriouk

>     1. I have suggested putting FormatPath or similar in ExpandString() override to make it easier.

I don't see how to do that or why. `FormatPath()` existed as a function in `ProjMgrYamlEmitter` before my changes, and the purpose as I understand it is for outputting useful paths in the YAML (not only for generator related paths). I'm not sure if you suggest that we move that functionality elsewhere. Ideally, the YAML emitter should get the arguments and any switches in a form that is as clean as possible, i.e. as separate strings and not joined already, and decide for itself how they should be output in the YAML format, or if and how any argument representing a path should be converted to be useful in the YAML output.

>     2. You must address quoting problems anyway since an argument element can contain something like:
>        `<argument>-I"$P/My Include Path Wish Spaces" </argument>`

I disagree. It's up to the consumer of the YAML output how spaces should be handled. For example a Python implementation would just hand the arguments as a list of strings to [subprocess.run()](https://docs.python.org/3/library/subprocess.html) and leave any spaces to be handled by that module. Quoting is necessary when generating a full command line to be consumed by a shell (i.e. in `GetExpandedCommandLine()`), but not in `GetExpandedArgv()`. Quoting arguments correctly and handling all corner cases (e.g. proper escaping of existing double and single quotes) is notoriously difficult, and I don't see why we even should try to do that in places where it also makes the output less usable.

edriouk:
@ErikMallbergSTM , then I would suggest stepping the iteration back:
-  take  the command line and arguments directly from RteItem instances provided by RteGenerator
-  modify RteGenerator methods to accept "host" argument

ErikMallbergSTM:
@edriouk

The changes I need are actually quite small, so I made them in a separate commit. Now all tests pass on my side.

edriouk (approved):
Looks good to me. 
Warnings about Construct() overload can be ignored, I'll address that later, already planned for RteModel refactoring

ErikMallbergSTM:
Thank you @edriouk and @brondani for the review!

brondani:
Thank you @ErikMallbergSTM for the contribution.

Issue #68, opened 2021-12-09
cbuildgen / make cmake default config files more generic with toolchain version


jeromecoutant:
Hi

Default cmake config files:
https://github.com/Open-CMSIS-Pack/devtools/tree/main/tools/buildmgr/cbuildgen/config

For ex, I am with GCC 10.3.1, but `GCC.10.2.1.cmake` is used.
It seems that some parts will not be correct:
https://github.com/Open-CMSIS-Pack/devtools/blob/main/tools/buildmgr/cbuildgen/config/GCC.10.2.1.cmake#L156-L158

We can't know which version has all users... How can we make these file more generic ?

Thx

brondani:
By design each `cmake` file in this directory should reflect what is installed in the environment. Different compiler versions can potentially have differences impacting this mapping.

In my opinion CMSIS-Build needs a toolchain discover and register utility.
It could enclose generic templates for the supported toolchain version ranges.

Such utility could automatically recognize compatible toolchains for example by searching them in the system `PATH` and in the most common system directories, check their versions and if compatible generate/register them in `$CMSIS_COMPILER_ROOT`.
It could also take as argument one or more specified directories to register or unregister toolchains contained in it.

@jeromecoutant What do you think about it?

jeromecoutant:
> In my opinion CMSIS-Build needs a toolchain discover and register utility.

Yes, install procedure in CMSIS toolbox should include this...

ReinhardKeil:
@jeromecoutant I suggest to close this issue.  We understand that we should work on a better installation process.

PR #679, opened 2023-01-23
[buildmgr] Add Star-MC1, Cortex-M55 and Cortex-M85 to IAR Compiler configuration.

grasci-arm:


JonatanAntoni:
@tarek-bochkati, please review this.

tarek-bochkati:
LGTM, but IMO we need to rename this cmake file to the latest IAR version supporting Cortex-M85 (9.30.1 or later)

@mario-pi

JonatanAntoni:
@tarek-bochkati, good catch. What versions of compilers shall be supported? Do we need multiple versions then?
cc @brondani

tarek-bochkati:
checking the history of GCC and ACx cmake files, I found that these files have been renamed to mirror the latest toolchain available versions.

also, I have checked that:
* **Cortex-M55** is supported since **IAR 9.20.4**
* **Cortex-M55** is supported since **IAR 9.30.1**

latest **IAR 9.32.1** is just adding Cortex-R82 support

Based on this, I recommend switching the latest 9.32.1

cc: @mario-pi

mario-pi:
@grasci-arm @JonatanAntoni thanks for providing this. As @tarek-bochkati noted, the file should be renamed to match the version of the IAR toolchain that supports the new cores. @tarek-bochkati should there be only one file for the latest greatest supported version, or is it possible to address multiple versions of the same toolchain? We are working to be able to test and provide the cmake files for IAR ourselves but unfortunately not up to speed yet.

brondani:
> @brondani, any thoughts about this? What happens with users of version 8 once we rename the file?

I wouldn't start shipping a separate file for every supported version due to the significant effort it could demand when it scales up. I believe we need a better strategy for toolchains configuration as commented [here](https://github.com/Open-CMSIS-Pack/devtools/issues/68#issuecomment-989938770).
For the time being I would suggest to ship only one file for the latest supported version.
In the current setup users need to manually update their configuration when [installing the tools](https://github.com/Open-CMSIS-Pack/cmsis-toolbox/blob/main/docs/installation.md) regardless of the toolchain version.

brondani (approved):

PR #734, opened 2023-02-10
[buildmgr] Fix projects using preincludes in IAR

JonatanAntoni:
Projects may fail with compiler error:
```Unexpected command line arguments: -include```
The correct compiler flag is `--preinclude`


JonatanAntoni:
@jlonnberg, @tarek-bochkati, may I ask you to review this, please?

jlonnberg:
@JonatanAntoni Looks ok for me. The same flag is valid for the 8 series of iccarm as well, so it might be a good idea to patch the older toolchain file as well

JonatanAntoni:
@jlonnberg, we only keep the config for the latest compiler version, here. If basically the same settings hold true for older compilers, the user can simply duplicate the file for multiple version, if required. Maintaining the toolchain configuration in an easy way is still on our backlog.

brondani (approved):

Issue #506, opened 2022-10-04
[packchk] embed XML validation as an initial validation step


jkrech:
This is a request to integrate an XML schema validation step in the `packchk` utility to avoid requiring the installation/presence of additional tools for the automated as well as manual pack creation and validation flows.
From experience we find that people miss to validate the schema compliance but fully rely on `packchk` instead. That tool however is limited to semantical validation and does not cover schema validation.

The schema validation should be run as a first step when invoking `packchk`. If the validation fails the program shall print out all available messages and bail out without running any semantic checks.

Consider command line options:
a) disable schema validation (default is "enabled")
b) option to specify the path to the PACK.xsd schema file

fred-r:
@jeromecoutant : would you like to volunteer for implementing this ?

jeromecoutant:
@fred-r I support the idea of this issue, but I can't implement :-)

fred-r:
> @fred-r I support the idea of this issue, but I can't implement :-)

@jkrech : is it really mandatory to merge the tools with new CLI options ?
May we keep the tools separated but provide a wrapper ?

This wrapper would be the new entry point to have both checks systematically made ?

jkrech:
I am not saying it is mandatory, however from my experience, asking people to install xmllint (in particular on Windows) as another prerequisite is a hurdle if not a barrier and I am not in favor re-distributing tools and utilities but in this case prefer to integrate some existing XML-Library to run the schema check from within `packchck` without creating an additional dependency. In csolution we do the same for checking the yml files while in cbuild we rely on a the command line option `-s` to include a schema check. In case xmllint is not found in the path you see this output.
```
cbuild -s "EventStatistic.Debug+AVH.cprj"
info cbuild: Build Invocation 1.1.0 (C) 2022 ARM
warning cbuild: xmllint was not found, proceed without xml validation
```

jkrech:
@yvesll , @tcsunhao you kindly agreed to contribute this feature to packchk. Please let me know if you have any questions.

yvesll:
@jkrech, in order to add a copy of the PACK.xsd in the repo, I have a discussion with @soumeh01. Sourabh suggests I track the tags of the file to make it flexible to manage the version of the file. So marking the open cmsis pack spec repo as an external submodule seems convenient as we can directly get any version of PACK.xsd. Or a more lightweight proposal is downloading the file from GitHub in runtime. 
Does anyone have different ideas or concerns?

[...]

PR #736, opened 2023-02-13
[packchk] Add xml validation for input PDSCs (#506)

yvesll:
Signed-off-by: yvesll <toyears@hotmail.com>

**NOTE:
This PR did not include the mandatory option which moves PACK.xsd to <cmsis_toolbox>/etc, I do not know the deployment flow.**

### DIFF
--- a/libs/xmlschemachecker/CMakeLists.txt
+++ b/libs/xmlschemachecker/CMakeLists.txt
@@ -1,0 +1,22 @@
+project(XmlValidator VERSION 1.0.0)
+
+add_subdirectory("test")
+
+SET(SOURCE_FILES XmlErrorHandler.cpp XmlValidator.cpp XmlErrorHandler.h XmlValidator.h XmlChecker.cpp)
+SET(HEADER_FILES XmlChecker.h)
+
+list(TRANSFORM SOURCE_FILES PREPEND src/)
+list(TRANSFORM HEADER_FILES PREPEND include/)
+
+add_library(XmlValidator STATIC ${SOURCE_FILES} ${HEADER_FILES})
+
+set_property(TARGET XmlValidator PROPERTY
+        MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
+
+target_include_directories(XmlValidator
+        PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include
+        PRIVATE ${CMAKE_BINARY_DIR}/external/xerces-c/src
+        PRIVATE ${CMAKE_SOURCE_DIR}/external/xerces-c/src
+        )
+
+target_link_libraries(XmlValidator PUBLIC ErrLog PRIVATE xerces-c)
--- a/tools/packchk/include/CreateModel.h
+++ b/tools/packchk/include/CreateModel.h
@@ -26,11 +26,14 @@ public:
   bool CheckForOtherPdscFiles(const std::string& pdscFullPath);
   bool AddPdsc(const std::string& pdscFile, bool bSkipCheckForOtherPdsc = false);
   bool AddRefPdsc(const std::set<std::string>& pdscRefFiles);
+  bool SetPackXsd(const std::string& packXsdFile);
   bool ReadAllPdsc();
   bool PrintPdscFiles(std::list<std::string>& pdscFiles);
 
 private:
   RteModelReader m_reader;
+  std::string m_schemaFile;
+  bool m_validatePdsc = false;
 
 };
 
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@
 *.opendb
 .vscode
 /build*/
+**/cmake-*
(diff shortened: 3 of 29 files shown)


JonatanAntoni:
Thanks @yvesll for your contribution. Unfortunately, I see some challenged with this enhancement. 

First of all, I'd like to understand why you think the schema check should be integrated into PackChk, natively.

I added an explicit schema check into the gen-pack script library, see https://github.com/Open-CMSIS-Pack/gen-pack/blob/27feb09121b49fdc7bf95e6927df9a50454cdffb/gen-pack#L80-L121. This uses xmllint from libxml2 which is available under MIT license for all supported platforms.

If you still think the schema check should be integral part of packchk, we'd need to sort out open questions. The PR currently fails the checks. This could be due to linking Xerces library dynamically but not shipping the required library alongside the binary. So far, we link everything statically and accept larger binaries to prevent shared lib issues across different platforms and versions.

Cheers,
Jonatan

yvesll:
HI, @JonatanAntoni, thanks for the comments.
For the failed checks, I am still exploring an elegant way to dynamically link the Xerces-c library and make sure it can work for all supported arches. Building the static lib may evidently increase the binary size but I appreciate it if we accept it.

For the motivation of this PR, you can see discussions in #506.

JonatanAntoni:
@yvesll, thanks for clarifying this. I think we should accept the static linkage overhead as we want our CMSIS-Toolbox to be independent from a specific Linux distro. Perhaps changing the build to use static link would already solve the remaining issue.

tools/packchk/include/CreateModel.h
```diff
@@ -10,6 +10,7 @@
 #include "Validate.h"
 #include "RteModelReader.h"
 #include "PackChk.h"
+#include "XmlValidator.h"
```
edriouk, line 13: XmlValidator includes Xerces includes, they should not poison the code.

tools/packchk/include/CreateModel.h
```diff
@@ -26,11 +27,15 @@ class CreateModel
   bool CheckForOtherPdscFiles(const std::string& pdscFullPath);
   bool AddPdsc(const std::string& pdscFile, bool bSkipCheckForOtherPdsc = false);
   bool AddRefPdsc(const std::set<std::string>& pdscRefFiles);
+  bool SetPackXsd(const std::string& packXsdFile);
   bool ReadAllPdsc();
   bool PrintPdscFiles(std::list<std::string>& pdscFiles);
 
 private:
   RteModelReader m_reader;
+  XmlValidator m_xmlValidator;
```
edriouk, line 36: Making this a dynamically allocated pointer would solve two problems:
1) no need to include XmlValidator.h in the header, forward declaration is sufficient.
2) the object can be created only if needed (validation is enabled)

libs/xmlschemachecker/test/data/PACK.xsd
```diff
@@ -0,0 +1,1980 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
edriouk, line 1: Could we pull this file from its original location?
https://github.com/Open-CMSIS-Pack/Open-CMSIS-Pack-Spec/tree/main/schema

libs/xmlschemachecker/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+list(TRANSFORM HEADER_FILES PREPEND include/)
+
+add_library(XmlValidator STATIC ${SOURCE_FILES} ${HEADER_FILES})
+
+set_property(TARGET XmlValidator PROPERTY
+        MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
+
+target_include_directories(XmlValidator
+        PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include/
+        SYSTEM ${CMAKE_BINARY_DIR}/external/xerces-c
```
edriouk, line 18: all xerces includes should be made private and not exposed to outside the library.

edriouk (changes requested):

yvesll:
Can someone help me rerun the "CodeQL" check? I do not have access and it seems a network issue.

yvesll:
@thorstendb-ARM, sorry, I click the re-request button but not sure why it removes you from the list, can you join the review again?

edriouk (approved):
Looks good.
@JonatanAntoni, could we disable static analysis for all external modules?

edriouk:
Code QL issues are ignored because they mostly come from an external  module Xerces

thorstendb-ARM:
@yvesll , why did you add this? I am trying to find out which is better, stopping after the first found error vs. reporting all errors including subsequent ones.
`m_domParser->setValidationConstraintFatal(true);`

yvesll:
@thorstendb-ARM, this field was set to test **setExitOnFirstFatalError** in the development stage and I forgot to remove it. I think both of the two ways make sense in different use cases, so an extra option is better.

PR #737, opened 2023-02-13
Clang 15 warnings

ErikMallbergSTM:
Fixes miscellaneous warnings when building with Clang 15.

Signed-off-by: Erik MÅLLBERG <erik.lundin@st.com>

### DIFF
--- a/libs/crossplatform/CMakeLists.txt
+++ b/libs/crossplatform/CMakeLists.txt
@@ -31,13 +31,6 @@ target_include_directories(CrossPlatform
   PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include
   PRIVATE ./src/${CMAKE_SYSTEM_NAME})
 
-if(NOT EMSCRIPTEN AND
-   ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" OR
-    "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang"))
-# Add library to support std::filesystem
-  target_link_libraries(CrossPlatform PUBLIC stdc++fs)
-endif()
-
 if(UNIX)
   target_include_directories(CrossPlatform PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include/win)
 elseif(WIN32)
--- a/libs/crossplatform/test/CMakeLists.txt
+++ b/libs/crossplatform/test/CMakeLists.txt
@@ -42,11 +42,6 @@ add_dependencies(CrossPlatformUnitTests TestProg)
 
 add_definitions(-DTEST_BIN_PATH="$<TARGET_FILE:CrossPlatformUnitTests>")
 
-if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" OR
-   "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
-    target_link_libraries(CrossPlatformUnitTests PUBLIC stdc++fs)
-endif()
-
 target_link_libraries(CrossPlatformUnitTests PUBLIC gtest gtest_main CrossPlatform RteFsUtils)
 
 add_test(NAME CrossPlatformUnitTests
--- a/libs/rtefsutils/CMakeLists.txt
+++ b/libs/rtefsutils/CMakeLists.txt
@@ -15,12 +15,4 @@ set_property(TARGET RteFsUtils PROPERTY
 
 target_include_directories(RteFsUtils PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include)
 
-if(NOT EMSCRIPTEN AND
-   ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" OR
-    "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang"))
-# Add library to support std::filesystem
-  target_link_libraries(RteFsUtils PUBLIC RteUtils CrossPlatform stdc++fs)
-else()
-  target_link_libraries(RteFsUtils PUBLIC RteUtils CrossPlatform)
-endif()
-
+target_link_libraries(RteFsUtils PUBLIC RteUtils CrossPlatform)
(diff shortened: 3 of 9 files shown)


ErikMallbergSTM:
I think it would be good to not squash commits when merging, to keep matters separated.

ErikMallbergSTM:
@edriouk and @brondani, could somebody please review?

edriouk (approved):

libs/crossplatform/include/CrossPlatform.h
```diff
@@ -9,7 +9,7 @@
 // Platform-specific API,  C-runtime methods and includes
 #ifdef _WIN32
 
-#include <Windows.h>
+#include <windows.h>
```
edriouk, line 12: Windows.h with capital W is correct

tools/svdconv/SVDGenerator/src/SfrccInterface.cpp
```diff
@@ -10,7 +10,7 @@
 #include "RteFsUtils.h"
 
 #ifdef _WIN32
-#include <Shlwapi.h>
+#include <shlwapi.h>
```
edriouk, line 13: Shlwapi.h with capital S is correct

libs/crossplatform/include/CrossPlatform.h
ErikMallbergSTM: It might be correct in the sense that it's capital 'W' in newer Windows SDKs, but since Windows is case-insensitive anyway, it doesn't matter when compiling on Windows, but when cross compiling, most tools seem to expect lowercase 'w', see https://stackoverflow.com/a/15466951. Several cross-platform software projects seem to have encountered this as well (for example [spdlog](https://github.com/gabime/spdlog/pull/2495/files), [Apache Thrift](https://github.com/apache/thrift/pull/2649) and [Blender](https://projects.blender.org/blender/blender/issues/34320)) and more examples can easily be found where this is already used ([Incscape](https://gitlab.com/inkscape/inkscape/-/blob/master/src/inkview-main.cpp#L17), [ffmpeg](https://git.ffmpeg.org/gitweb/ffmpeg.git/blob/refs/heads/master:/compat/w32dlfcn.h#l25), [the Apache web server](https://svn.apache.org/viewvc/httpd/httpd/trunk/support/win32/wintty.c?revision=420983&view=markup#l38), [Electron](https://github.com/electron/electron/blob/main/shell/browser/browser_win.cc#L10)). So I don't think it's so much a matter of what is "correct" but if we would like to make it easier for anybody wanting to cross compile (and I don't see why we wouldn't).

edriouk: Do we see errors/warning with windows includes on other platforms? 
Do we use cross-compile? 
We do use the latest Windows SDK only.
Besides we have introduced substitutes for windows headers  in libs/crossplatform/include/win as safeguards. 
They must need to be renamed too since they are used in non-windows environments.
Therefore I would not change the includes in this PR and address that only when it become an issue..

ErikMallbergSTM: > Do we see errors/warning with windows includes on other platforms, including static analysis?

Without this fix I get this error when building for Windows on Linux:
```
[  1%] Building CXX object libs/crossplatform/CMakeFiles/CrossPlatform.dir/src/Windows/Utils.cpp.obj
[...]/devtools/libs/crossplatform/src/Windows/Utils.cpp:10:10: fatal error: 'Windows.h' file not found
#include <Windows.h>
         ^~~~~~~~~~~
1 error generated.
make[2]: *** [libs/crossplatform/CMakeFiles/CrossPlatform.dir/build.make:92: libs/crossplatform/CMakeFiles/CrossPlatform.dir/src/Windows/Utils.cpp.obj] Error 1
make[1]: *** [CMakeFiles/Makefile2:970: libs/crossplatform/CMakeFiles/CrossPlatform.dir/all] Error 2
make: *** [Makefile:146: all] Error 2
```

> Do we use cross-compile?

It's not about what we do but how portable we would like the code to be. I find it convenient to be able to cross compile for Windows without leaving my main Linux machine.

> We do use the latest Windows SDK only. Besides we have introduced substitutes for windows headers in libs/crossplatform/include/win as safeguards. They must need to be renamed too since they are used in non-windows environments. Therefore I would not change the includes in this PR and address that only when it become an issue..

I don't think I follow. Do the changes in this PR stop your builds from passing?

edriouk: This PR addresses Clang warnings. Therefore I do not think changes in include filenames are relevant to it.

Which compiler do you use for cross-compilation? We currently do not support GCC on Windows, see README.md: 
☑️ GCC/Clang on Windows:
    Currently GCC and Clang (MSYS2/MinGW distribution) compilers do not work
    on Windows. The included libc++ has a known issue in std::filesystem,
    see [MSYS2 GitHub issue #1937](https://github.com/msys2/MSYS2-packages/issues/1937).

ErikMallbergSTM: > This PR addresses Clang warnings. Therefore I do not think changes in include filenames are relevant to it.

I'm fine moving this change to a new PR.

> Which compiler do you use for cross-compilation? We currently do not support GCC on Windows, see README.md: ballot_box_with_check GCC/Clang on Windows: Currently GCC and Clang (MSYS2/MinGW distribution) compilers do not work on Windows. The included libc++ has a known issue in std::filesystem, see [MSYS2 GitHub issue #1937](https://github.com/msys2/MSYS2-packages/issues/1937).

For cross compilation, I'm using Clang 15.0.7, included with the current development version of [Zig](https://ziglang.org/). The referenced bug didn't manifest when I cross compiled the snippet in the bug report and executed on windows (the destination file was overwritten as expected).

edriouk: Hi Erik,
@JonatanAntoni , @brondani and me have had a discussion and came to conclusion that your suggestion to use lowercase in the windows includes is good as it brings advantages like cross-compile. 
However please create a separate PR for Windows includes.

ErikMallbergSTM: @edriouk Great! Those changes now moved to #743.

edriouk (approved):

PR #761, opened 2023-02-22
GitHub: Don't run workflows for changes in docs

JonatanAntoni:


tools/projmgr/docs/Template/MyProject/.clangd
```diff
@@ -0,0 +1,3 @@
+CompileFlags:
```
jkrech, line 1: this file reference contains your personal file location. Is this useful to be checked in?

jkrech (approved):
Looks good to me.

PR #763, opened 2023-02-23
GitHub: Cache gcc-arm installer

JonatanAntoni:
The gcc-arm installer archive is huge. To prevent downloading the file for each run it is cached in GitHub based on the download url cache-key.

### DIFF
--- a/.github/workflows/buildmgr.yml
+++ b/.github/workflows/buildmgr.yml
@@ -115,7 +115,7 @@ jobs:
 
       - name: Install deps
         run: |
-          sudo apt update
+          sudo apt-get update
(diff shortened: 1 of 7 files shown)


.github/workflows/buildmgr.yml
```diff
@@ -480,11 +480,17 @@ jobs:
       - name: Create build folder
         run: mkdir build
 
+      - name: Cache ARM GCC
+        id: cache-arm-gcc
```
soumeh01, line 484: Since we are not using `id`.  We can remove it.

soumeh01 (dismissed):

soumeh01 (approved):

PR #837, opened 2023-04-14
Fix branch protection examples

grasci-arm:


jkrech (approved):
Looks sensible to me to assure that the project do not build with old versions of the compiler.

PR #877, opened 2023-05-03
[projmgr] Add list environment command

grasci-arm:


tools/projmgr/src/ProjMgr.cpp
```diff
@@ -672,6 +686,17 @@ bool ProjMgr::RunListToolchains(void) {
   return ret;
 }
 
+bool ProjMgr::RunListEnvironment(void) {
+  EnvironmentList env;
+  if (!m_worker.ListEnvironment(env)) {
+    ProjMgrLogger::Error("processing configurations list failed");
+    return false;
```
brondani, line 693: @soumeh01 It seems this is not being covered: `ProjMgrWorker::ListEnvironment` never returns `false`.

brondani (approved):

Issue #559, opened 2022-10-26
`RTE_components.h`, preinclude files generated in wrong location and too late


ReinhardKeil:
Currently the file `RTE_components.h` and preinclude files are generated by cbuild in the directory with this naming:

```
$ProjectDir$/RTE/_$Project$.$Build-Type$_%TargetType$
```

I suggest that these files are generated by `csolution` and the directory is changed to:
```
$ProjectDir$/RTE/_$Build-Type$_%TargetType$
```

The change of the directory makes the behavior identical to MDK (when a *.CPRJ is imported)
The change to generate the files by `csolution` allows to run static code analysis tools or intellisense engines before a build (these tools need visibility to all #define symbols).

brondani:
The creation of the RTE folder content, including RTE_Components.h and pre-include files, the introduction of flags to optionally skip it in `csolution` and the inverted logic in `cbuild` are being implemented in PR https://github.com/Open-CMSIS-Pack/devtools/pull/432 according to issue https://github.com/Open-CMSIS-Pack/devtools/issues/371.

The target subdirectory containing only `build-type` and `target-type` is not unique when multiple projects have the same root or when they occasionally set the same custom RTE folder.

ReinhardKeil:
>The target subdirectory containing only build-type and target-type is not unique when multiple projects have the same root or when they occasionally set the same custom RTE folder.

Is this really a valid use case?  When would multiple projects share the same RTE directory? Should we make this directory configurable instead?

brondani:
Unless we forbid multiple cproject.yml files in the same directory, it is a valid use case. CoreValidation already has such structure:
https://github.com/ARM-software/CMSIS_5/tree/develop/CMSIS/CoreValidation/Project

This directory could follow the configurable `target-name` as in your proposal https://github.com/Open-CMSIS-Pack/devtools/issues/558.

ReinhardKeil:
Thanks for pointing to the use case, I will review it.

PR #909, opened 2023-05-09
PackChk: improved error reporting for invalid command line and when PACK.XSD has not been found. (#559)

grasci-arm:
Also checked M313 against mapped drive on windows, e.g. "subst D: C:\D" (for path "C:\D")


jkrech (approved):
@thorstendb-ARM please work with @soumeh01 to ensure that the most recent version of the PACK.xsd is included in the cmsis-toolbox release in the `./etc` directory.

PR #975, opened 2023-06-09
Create *.cbuild-gen.yml with absolute paths

ErikMallbergSTM:
To facilitate implementation of generators, create *.cbuild-gen.yml in addition to *.cbuild.yml with any file and directory paths inside expanded to absolute paths.

The *.cbuild-gen.yml is created inside the intdir, since this resolved version is only a temporary file used by generators.

The $G parameter for generators now expands to the *.cbuild-gen.yml file.

### DIFF
--- a/tools/projmgr/src/ProjMgrYamlEmitter.cpp
+++ b/tools/projmgr/src/ProjMgrYamlEmitter.cpp
@@ -30,7 +30,7 @@ class ProjMgrYamlCbuild {
 private:
   friend class ProjMgrYamlEmitter;
   ProjMgrYamlCbuild(YAML::Node node, const vector<ContextItem*> processedContexts, ProjMgrParser& parser, const string& directory);
-  ProjMgrYamlCbuild(YAML::Node node, const ContextItem* context);
+  ProjMgrYamlCbuild(YAML::Node node, const ContextItem* context, bool useAbsolutePaths = false);
   void SetContextNode(YAML::Node node, const ContextItem* context);
   void SetComponentsNode(YAML::Node node, const ContextItem* context);
   void SetComponentFilesNode(YAML::Node node, const ContextItem* context, const string& componentId);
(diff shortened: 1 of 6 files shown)


brondani (commented):
For the sake of solutions' portability I would recommend to avoid absolute paths as default behaviour.
My suggestion in this case would be to add a dedicated command for the sole purpose of generating the "generator input file" - which could have a different file extension - and a command line option for generating absolute paths instead of the default relative paths.

jkrech:
This PR is on hold!
Let's not merge this PR prior to cmsis-toolbox 2.0.0 release. 
We are currently in discussion how we move forward with cbuild-idx.yml cbuild.yml files as we see multiple use cases:
a) a log file checked-into a repository to detect differences when re-building or after changes
b) a lock file which specifies versions of packs and components being used where they are not fixed in csolution, cproject or clayer.
c) a generator input file

jkrech:
Instead of replicating all cbuild.resolved.yml files for all converted contexts, I propose to have the csolution run command for generators generating a dedicated generator cbuild formatted file prior to the invokation of the generator including the packID + generatorID and absolute file paths for easing the processing by the generators. The life span of this file is short and we could consider removing the file after the generator has completed its operation.

jkrech:
As discussed in the last technical project meeting the idea was to only generate a `<context>.cbuild-gen.yml` as part of the 
```
csolution run <csolutionname>.csolution.yml -c <context> -g <generatorID> [-p <packID>]
```
command. So prior to invoking the generator this file is created.

The simple solution is to place the <context>.cbuild-gen.yml side by side with the <context>.cbuild.yml.
If the generator invokation includes $G it could be located in the `<intdir>` for the context. Since the cbuild-gen.yml provides absolute paths only the location does not matter for generator.

Am I missing something?

ErikMallbergSTM:
Superseded by #1129

PR #976, opened 2023-06-12
Update to xerces-c b597a0f

Torbjorn-Svensson:
This fixes cross compile issue for Windows build on Linux.


jkrech:
Looks good to me, since windows is case insensitive this change does not impact the windows build on windows but according to @Torbjorn-Svensson it needs to be lower case in case of building on a linux machine.
@thorstendb-ARM  could you please take a look as well.

Torbjorn-Svensson:
There is one action that fails, but I'm certain that the case change of a header file in xerces-c should not trigger a failure to upload to codecov.io.

jkrech:
Unfortunately the code coverage service is not reliable and I had just re-run that successfully.

jkrech (approved):
Looks good to me.

PR #985, opened 2023-06-15
Unit test to reproduce crash in ProjMgrYamlEmitter

### LINKED ISSUE #986: csolution crashes on Mac on project file name case mismatch
csolution crashes on Mac if the filename case used in `csolution.yml` doesn't match the actual filename case on the disk.

JonatanAntoni:
File name case issue causes `std::length_error` in [ProjMgrYamlEmitter.cpp:75](https://github.com/Open-CMSIS-Pack/devtools/blob/107f88e83b1416ce3896e48519a87e689767ab9a/tools/projmgr/src/ProjMgrYamlEmitter.cpp#L71-L75). The lookup `std::find_if` returns end iter due to none matching file name case.

Fixes #986


tools/projmgr/test/src/ProjMgrUnitTests.cpp
```diff
@@ -3351,3 +3351,22 @@ solution:\n\
...
+* that references a project using different file name case.
+*/
+TEST_F(ProjMgrUnitTests, RunProjMgr_YamlEmitterFileCaseIssue) {
+  char* argv[7];
+  const string& csolution = testinput_folder + "/TestSolution/FilenameCase/filenbame.csolution.yml";
+  argv[1] = (char*)"convert";
+  argv[2] = (char*)csolution.c_str();
+  argv[5] = (char*)"-o";
+  argv[6] = (char*)testoutput_folder.c_str();
+  EXPECT_EQ(0, RunProjMgr(7, argv, 0));
```
brondani, line 3365:
```suggestion
  EXPECT_EQ(1, RunProjMgr(7, argv, 0));
```

If we want to make sure the projects are portable I believe we should throw an error in this case, we should discuss about it.

tools/projmgr/test/src/ProjMgrUnitTests.cpp
```diff
@@ -3351,3 +3351,22 @@ solution:\n\
   }
 }
 
+/*
+* This test case runs the ProjMgrYamlEmitter with a solution
+* that references a project using different file name case.
+*/
+TEST_F(ProjMgrUnitTests, RunProjMgr_YamlEmitterFileCaseIssue) {
+  char* argv[7];
+  const string& csolution = testinput_folder + "/TestSolution/FilenameCase/filenbame.csolution.yml";
```
brondani, line 3360:
```suggestion
  const string& csolution = testinput_folder + "/TestSolution/FilenameCase/filename.csolution.yml";
```

JonatanAntoni: Typo fixed.

tools/projmgr/test/src/ProjMgrUnitTests.cpp
JonatanAntoni: Yes, we need to discuss if we accept different file name case. Otherwise we need to detect this as an issue in validation phase and stop gracefully.

jkrech: We agreed to not allow mismatch in file name cases since we aim for portability.

brondani:
> The fix doesn't work on Linux where the file with wrong case is considered non-existing. Hence the test case triggers the first if-clause already. While this has technically the same effect, the test case doesn't receive the expected error message.

You could use `CrossPlatfomUtils::GetHostType()` to set different test expectations depending on the platform.

tools/projmgr/src/ProjMgr.cpp
```diff
@@ -380,6 +380,10 @@ bool ProjMgr::PopulateContexts(void) {
       if (cprojectFile.empty()) {
         ProjMgrLogger::Error(cproject, "cproject file was not found");
         return false;
+      } else if (fs::path(cprojectFile).filename().string() != fs::path(cproject).filename().string()) {
+        ProjMgrLogger::Error(cproject, "cproject filename has case inconsistency");
+        ProjMgrLogger::Info(fs::path(cprojectFile).filename().string(), "should be used instead");
```
brondani, line 385:
```suggestion
        ProjMgrLogger::Info(fs::path(cprojectFile).filename().generic_string(), "should be used instead");
```
`generic_string()` prints consistently the same directory separator `/` for all platforms.

JonatanAntoni: In this case only the `filename()` part is considered so there should not be any directory separators involved.

JonatanAntoni:
The failing test case is due to strict case sensitivity on Linux. While on Windows and Mac a `fs::path("path/to/file")` would pick up `path/to/File` and report this as a case-mismatch, on Linux this would *not* consider the `File` and rather report this as `path/to/file does not exist`. 

This results in different error messages depending on the platform. If we want to accept this, the test case assertion needs to differentiate.

JonatanAntoni:
> ## [Codecov](https://app.codecov.io/gh/Open-CMSIS-Pack/devtools/pull/985?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Open-CMSIS-Pack) Report
> > Merging [#985](https://app.codecov.io/gh/Open-CMSIS-Pack/devtools/pull/985?src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Open-CMSIS-Pack) ([51cf298](https://github.com/Open-CMSIS-Pack/devtools/commit/51cf298db3cfda5784cb5cc49604ea256a435f34)) into [main](https://app.codecov.io/gh/Open-CMSIS-Pack/devtools/commit/9d3d104b937f60a4df71d3a45307cf5d1875e785?el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Open-CMSIS-Pack) ([9d3d104](https://github.com/Open-CMSIS-Pack/devtools/commit/9d3d104b937f60a4df71d3a45307cf5d1875e785)) will **decrease** coverage by `0.01%`.
> > The diff coverage is `0.00%`.
> 
> Additional details and impacted files

This is expected as we introduced "platform dependent code" but collecting coverage on Linux only.

brondani (approved):
LGTM

PR #988, opened 2023-06-15
Update gcc_linker_script.ld

jkrech:
missing start and length for bss (1) in zero table.


jkrech:
I now learned that this is not a defect but done intentionally.

The first version of the GCC linker scripts added the bss start and end labels to the zero table. But during debugging it turned out that bss section was zero initialized twice. One time via the the zero table initialization in the CMSIS startup code and one time via the GCC library initialization. So we skip the 1st initialization.

PR #992, opened 2023-06-18
Define toolchain for cross build win32

Torbjorn-Svensson:
This is the toolchain file that I used in pr #976.


JonatanAntoni:
Sounds basically fine to me. How is this file being picked up by CMake?
I'd suggest to enhance the documentation about how to cross compiler on Windows for Linux.

PR #993, opened 2023-06-18
Directories should be executable even if read-only

Torbjorn-Svensson:
If directories are not marked as executable, it's not possible to list the content. This greatly complicates navigating the "output" directory tree when running the test suite.


JonatanAntoni (commented):
Sounds good to me.
Do we need to check Go-based tools, like cpackget and cbuild, for similar issues? These don't rely on RteFsUtils.

brondani:
At least the test case `RteChkTest.Summary` in `RteModelUnitTests` is failing.

JonatanAntoni:
@Torbjorn-Svensson, thanks for pointing this out. We reviewed the implementation of `SetFileReadOnly` and `SetTreeReadOnly` functions and think both have issues we should address, properly.

First of all, `SetFileReadOnly` suggests to handle files only. So using the function for directories is a bit misleading. But we could accept this. The major issue seems to be that we mess with all the permission flags, not only with the write flags. The worst part is that we make files and directories world-writeable. This could be considered a security issue. Hence, I'd suggest to read the current permissions and remove the write flags, specifically, leaving all other permission flags as they are.

Restoring write flags this might be a bit more complicated. How to know which of the three write flags to set? The owner one? I'd say yes. The group one? I'd say this depends on the group's read flag. And the worlds one? I'd say probably never. But there might be rare cases where one would want the files/dirs to be world writeable. We probably can't know.

Another issue might be burrowed in `SetTreeReadOnly`. We make the top level directory writeable to be able to change the permission flags of the content. Is this required? If yes, what about recursive sub-directories? We do not touch their flags.

I propose to create a couple of test cases, first. Once we have test cases for different situations we want to functions to perform in, we can focus on a proper fix.

Torbjorn-Svensson:
> @Torbjorn-Svensson, thanks for pointing this out. We reviewed the implementation of `SetFileReadOnly` and `SetTreeReadOnly` functions and think both have issues we should address, properly.
> 
> First of all, `SetFileReadOnly` suggests to handle files only. So using the function for directories is a bit misleading. But we could accept this. The major issue seems to be that we mess with all the permission flags, not only with the write flags. The worst part is that we make files and directories world-writeable. This could be considered a security issue. Hence, I'd suggest to read the current permissions and remove the write flags, specifically, leaving all other permission flags as they are.
> 
> Restoring write flags this might be a bit more complicated. How to know which of the three write flags to set? The owner one? I'd say yes. The group one? I'd say this depends on the group's read flag. And the worlds one? I'd say probably never. But there might be rare cases where one would want the files/dirs to be world writeable. We probably can't know.

One way to solve this would be to keep a map between the path and the original permission. On each change, the current permission and the path is added to the map. Then for each restore operation, the permission is fetched from the map and applied to the file and then removed from the map.
This approach would allow the change of permission without being unable to restore to the correct permission, but the down side is that it will consume memory and might seem complex.
How should a path be handled if it's not part of the map? Should the new permission be `0777 - umask`?

> Another issue might be burrowed in `SetTreeReadOnly`. We make the top level directory writeable to be able to change the permission flags of the content. Is this required? If yes, what about recursive sub-directories? We do not touch their flags.

To my knowledge, you are still allowed to overwrite the content of a file in a directory that is read-only, but you are not allowed to remove or add new entries in the directory.
I don't know why the permissions was not changed recursively if the files were supposed to be read only. Maybe an oversight in the original implementation?

> I propose to create a couple of test cases, first. Once we have test cases for different situations we want to functions to perform in, we can focus on a proper fix.

Okay, How do we want this to work then?
I suppose we want to recursively remove the write permission for all files or directories when requesting read-only tree.

> 
> Would you take this effort on? If not, I'd ask you to create an bug issue we can use for tracking, please.

Lets conclude on how we want the behavior to be fist.


On a side note, it looks like the `RteFsUtils::SetTreeReadOnly` function is only used by the test suite, so maybe not a major security issue right now.

Torbjorn-Svensson:
> At least the test case `RteChkTest.Summary` in `RteModelUnitTests` is failing. https://github.com/Open-CMSIS-Pack/devtools/actions/runs/5304401934/jobs/9615751340#step:6:274

I think these failure might be related to that it's now possible to recursively scan the content of the directories and more packs are picked up, but I'm not sure.
Let's conclude on how we should handle the read-only thing before trying to understand why the result of the test case is different.

libs/rtefsutils/src/RteFsUtils.cpp
```diff
@@ -299,22 +301,19 @@ bool RteFsUtils::SetTreeReadOnly(const string& path, bool bReadOnly) {
...
+  // Set permission on directory
+  if (!SetFileReadOnly(path, bReadOnly)) {
     return false;
   }
-  // Iterate over file tree
+
+  // Iterate over tree and change permission of files and directories
   for(auto& p: fs::recursive_directory_iterator(path, ec)) {
-    if (fs::is_regular_file(p.path(), ec)) {
+    if (fs::is_regular_file(p.path(), ec) || fs::is_directory(p.path(), ec)) {
```
Torbjorn-Svensson, line 311: I'm a bit uncertain if we should limit to files or directories here or if we should update any other kind of resource too.
What's your opinion?

JonatanAntoni: Well, I think that could be debated. What type of paths do we expect? I'd think files, directories and symlinks in the first place. Lets turn the question around: What could go wrong if we remove the condition?

JonatanAntoni:
> What if the parent directory is not writable? Should the resource still be writable in this case? If so, by whom? Owner only?

Good point! Could we go with `umask`, i.e. thy systems default creation file mode? On my Mac `umask` is `022`, i.e. only owner has write permissions by default.

libs/rtefsutils/src/RteFsUtils.cpp
```diff
@@ -269,15 +269,17 @@ bool RteFsUtils::SetFileReadOnly(const string& path, bool bReadOnly) {
...
 
   error_code ec;
-  fs::perms perms;
+  const fs::perms write_mask = fs::perms::owner_write | fs::perms::group_write | fs::perms::others_write;
 
-  // Set file permissions
   if (bReadOnly) {
-    perms = fs::perms::owner_read | fs::perms::group_read | fs::perms::others_read;
+    // Remove write bits
+    fs::permissions(fs::path(path), write_mask, std::filesystem::perm_options::remove, ec);
```
JonatanAntoni, line 276:
```suggestion
    fs::permissions(fs::path(path), write_mask, fs::perm_options::remove | s::perm_options::nofollow, ec);
```

libs/rtefsutils/src/RteFsUtils.cpp
```diff
@@ -269,15 +269,17 @@ bool RteFsUtils::SetFileReadOnly(const string& path, bool bReadOnly) {
  */
 
   error_code ec;
-  fs::perms perms;
+  const fs::perms write_mask = fs::perms::owner_write | fs::perms::group_write | fs::perms::others_write;
```
JonatanAntoni, line 272: Make a real compile-time constant here:
```suggestion
  constexpr auto write_mask = fs::perms::owner_write | fs::perms::group_write | fs::perms::others_write;
```

libs/rtefsutils/src/RteFsUtils.cpp
Torbjorn-Svensson: I don't think you can change permission on a symlink, but I'm not sure if it's silent or if an error is raised.
For any other type of resource, I think you can change the permission on a Linux FS.

JonatanAntoni: I am sure you can. A symlink is kinda special, as you can either change the symlink itself or the element it links to. According to docs we need to add `fs::filesystem::perm_options::nofollow` in order to affect the symlink and not its target,

JonatanAntoni:
@brondani, @edriouk: Can we somehow estimate what parts are affected when we change these functions? Do we expect causing any issues which might not be covered by tests?

libs/rtefsutils/src/RteFsUtils.cpp
```diff
@@ -269,15 +270,41 @@ bool RteFsUtils::SetFileReadOnly(const string& path, bool bReadOnly) {
...
+    fs::perms perms = fs::status(fs::path(path).parent_path(), ec).permissions() & write_mask;
+
+    // If parent directory is not writable, lets rely on current umask
+    if (perms == fs::perms::none) {
+      // Only way to get current umask is to change it, so revert the possible change directly.
+      mode_t umask = ::umask(0);
+      ::umask(umask);
+
+      // Map the mode_t value to coresponding fs::perms value
+      const struct {
```
Torbjorn-Svensson: Should this also be a constexpr?

libs/rtefsutils/src/RteFsUtils.cpp
```diff
@@ -269,15 +270,41 @@ bool RteFsUtils::SetFileReadOnly(const string& path, bool bReadOnly) {
...
+    fs::permissions(fs::path(path), write_mask, fs::perm_options::remove | fs::perm_options::nofollow, ec);
   } else {
-    perms = fs::perms::all;
+    // Inherit write bits from parent directory
+    fs::perms perms = fs::status(fs::path(path).parent_path(), ec).permissions() & write_mask;
+
+    // If parent directory is not writable, lets rely on current umask
+    if (perms == fs::perms::none) {
+      // Only way to get current umask is to change it, so revert the possible change directly.
+      mode_t umask = ::umask(0);
```
Torbjorn-Svensson, line 285: It does not appear to be a way to retrieve the umask value without actually changing it.
Is it fine to do it like this or should we find a different way to get the value?

libs/rtefsutils/src/RteFsUtils.cpp
```diff
@@ -269,15 +270,41 @@ bool RteFsUtils::SetFileReadOnly(const string& path, bool bReadOnly) {
...
 
-  // Set file permissions
   if (bReadOnly) {
-    perms = fs::perms::owner_read | fs::perms::group_read | fs::perms::others_read;
+    // Remove write bits
+    fs::permissions(fs::path(path), write_mask, fs::perm_options::remove | fs::perm_options::nofollow, ec);
   } else {
-    perms = fs::perms::all;
+    // Inherit write bits from parent directory
+    fs::perms perms = fs::status(fs::path(path).parent_path(), ec).permissions() & write_mask;
```
Torbjorn-Svensson, line 280: Should the function return `false` if `fs::status` returns error?

libs/rtefsutils/src/RteFsUtils.cpp
```diff
@@ -269,15 +270,41 @@ bool RteFsUtils::SetFileReadOnly(const string& path, bool bReadOnly) {
...
+        {S_IWGRP, fs::perms::group_write},
+        {S_IWOTH, fs::perms::others_write}
+      };
+      for (auto [m, p] : mode_to_perm_map) {
+        if ((umask & m) == 0) {
+          perms |= p;
+        }
+      }
+    }
+
```
Torbjorn-Svensson, line 303: Should the function return `false` if perms is still `fs::perms::none` here?

libs/rtefsutils/src/RteFsUtils.cpp
JonatanAntoni: I raised this question with some C++ experts here. It looks weird but if it works we might keep it for the time being.

libs/rtefsutils/src/RteFsUtils.cpp
JonatanAntoni: Probably yes.

libs/rtefsutils/src/RteFsUtils.cpp
```diff
@@ -25,6 +25,7 @@
 #include <iomanip>
 #include <regex>
 #include <thread>
+#include <sys/stat.h>
```
JonatanAntoni, line 28: This is not platform independent and fails at least on Windows. We can't use `umask`, at least not directly.

JonatanAntoni: Well, okay. Not sure about that. The Windows builds are broken. Perhaps we need to add additional library to the project?

Torbjorn-Svensson: Appears that the Windows umask function does not behave like a normal UNIX umask, so maybe it's better to skip the umask fallback on Windows?

JonatanAntoni: Well, we need to abstract this functionality if we can't rely on the c/c++ libraries. As a coding guideline we have defined to not use platform dependent code (via `#ifdef`'s) anywhere than in the CrossPlatform lib. We'd need to add a `umask` function to `CrossPlatformUtils` and implement it for all platforms. For Windows we might use some static defaults if we can't retrieve the setting.

Torbjorn-Svensson: I created a function GetCurrentUmask() that returns the current umask as a std::filesystem::perms instance. For Windows, I mapped the bits to the owner, but perhaps it should be mapped to (owner | group | other).

Please take a look and see if this is better. :)

libs/crossplatform/src/Windows/Utils.cpp
```diff
@@ -133,4 +136,29 @@ CrossPlatformUtils::REG_STATUS CrossPlatformUtils::GetLongPathRegStatus()
   }
   return REG_STATUS::DISABLED;
 }
+
+std::filesystem::perms CrossPlatformUtils::GetCurrentUmask() {
+  // Only way to get current umask is to change it, so revert the possible change directly.
+  unsigned int umask = ::umask(0);
```
JonatanAntoni, line 142: `umask` is deprecated according to https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/posix-umask?view=msvc-160. It recommends to use `_umask` or `_umask_s` instead. As this implementation is now Windows specific, I think we should follow the Microsoft recommendation.

JonatanAntoni:
There are two failing test cases. One seems to have an issue on all platforms but I don't see why this should be caused by the change. The other test case fails only on Windows due to unexpected file permissions.

libs/rtefsutils/test/src/RteFsUtilsTest.cpp
```diff
@@ -727,33 +727,43 @@ TEST_F(RteFsUtilsTest, RemoveDirectoryAutoRetry) {
...
   RteFsUtils::CreateFile(filenameRegular, "foo");
 
+  // Set permission to a known value (0644), with out this, the tests below will fail.
+  fs::permissions(filenameRegular, fs::perms::owner_write | fs::perms::owner_read | fs::perms::group_read | fs::perms::others_read, ec);
+
   // Test filename with regular separators
   ret = RteFsUtils::SetFileReadOnly(filenameRegular, true);
   EXPECT_EQ(ret, true);
   EXPECT_EQ((fs::status(filenameRegular, ec).permissions() & write_mask), fs::perms::none);
+  EXPECT_EQ(fs::status(filenameRegular, ec).permissions() & exec_mask, fs::perms::none);
```
Torbjorn-Svensson, line 742: I've tried to figure out why this is not working on Windows and it appears that it might not be so strange after all.
Based on your comments @JonatanAntoni, I added the verification that the execute flag was removed, but this test will never work on Windows. See the [Mircorsoft STL implementation of `_Permissions`](https://github.com/microsoft/STL/blob/main/stl/inc/filesystem#L3974) for the reason.

Is it okay to just drop the check of on the exec bits or should I make the test conditional using ifdef?

JonatanAntoni: This test makes me wondering. I think we agreed that we won't touch any flags other than the write ones. We should never remove exec flags nor set them. Hence the check should be removed, anyway.

JonatanAntoni: Ah, sorry, I misunderstood your comment. You wanted to add a check that only the write flags got affected and no others. Fair point.

Instead of check the flags explicitly; would it work to store the current permission of the file right after creation? And then verify after `SetFileReadOnly` only the write flags are changed/removed? The status of all other flags should just be untouched.

JonatanAntoni:
The second failing test case (`RteCheckTest.Summary`) is caused by an older commit, see #1002.

libs/rtefsutils/src/RteFsUtils.cpp
Torbjorn-Svensson: What's you view on it @JonatanAntoni, is it okay like it is now?

Torbjorn-Svensson:
I do not intend to address more of the permission changes mentioned in #1006 in this PR.
The remaining test that fail is due to that it's unable to upload to codecov.io.

libs/rtefsutils/src/RteFsUtils.cpp
JonatanAntoni: I think its okay, even if one could consider the solution over-engineered.

Perhaps, instead of using `umask` we could simply fall back to owner-only as this is what we need to make changes. I think using owner-only on Windows is fine.

JonatanAntoni:
We have CMSIS-Toolbox 2.0.0 final release scheduled and we don't want to brake anything, easily. So, we might want to wait for @edriouk to give the change a final review, probably next week. Then we can decide to take the fix in or postpone it to 2.1.0 release. Is this okay with you?

libs/rtefsutils/src/RteFsUtils.cpp
Torbjorn-Svensson: So, do you want me to change it to be hard-coded `0700` on Windows and leave the platforms at `0777 & ~umask()` or should all of them be hard-coded or keep it as it is now?
I would really want to get this merged as it's supper annoying to run the test-suite locally when files are no longer accessible...

JonatanAntoni: Lets keep it as is for now and wait for @edriouk.

I appreciate the annoyance of this but in light of the planned 2.0.0 release I'd rather like to be super conservative. I feel the same on Mac here, because I need to run git clean through sudo.

If this is not an immediate bug blocking features in 2.0.0, I'd give us some more days to conclude properly.

Torbjorn-Svensson: Ok. Let's give @edriouk a few days to review and revisit the topic early next week.

I think the security implications with files "randomly" becoming world writable (#1006) is something that should be fixed before the 2.0.0 as it could be dangerous for customer projects.

JonatanAntoni: I agree about the security implication. Unfortunately, this fix doesn't address all the issues but only parts of it. And since pack installation is now handled by `cpackget` the fix in RteUtils don't have any impact.

We need to clarify in what situations these functions are effectively used in released tools.

edriouk (approved):
As discussed with @JonatanAntoni , this particular PR has a low risk and can be approved for the upcoming release. Fixes for other permission issues will be done only after 2.0.0

PR #994, opened 2023-06-18
Make value of "VERSION_PRE_RELEASE" optional

Torbjorn-Svensson:
When building a git commit that is not tagged, the pre-release part of the tag name is optional.

Example without this change, the VERSION_FULL symbol would be: `0.9.2-tools/packgen/0.9.2+p338-gd366fa6`
With the change, it would instead be: `0.9.2+p338-gd366fa6`

### DIFF
--- a/cmake/ProjectVersionFromGitTag.cmake
+++ b/cmake/ProjectVersionFromGitTag.cmake
@@ -39,7 +39,11 @@ function(get_version_from_git_tag _prefix)
         if(DESCRIBE MATCHES "(-[a-zA-Z0-9-\\.]+)(-[0-9]+-g[0-9a-f]+)?$")
             # Version with PreRelease info
             string(REGEX REPLACE "(-[0-9]+)(-g[0-9a-f]+)?$" "" TAG_SEGMENT_STR "${DESCRIBE}")
-            string(REGEX REPLACE "^${_prefix}[0-9]+\\.[0-9]+\\.[0-9]+-(.*)" "\\1" VERSION_PRE_RELEASE "${TAG_SEGMENT_STR}")
+            string(REGEX REPLACE "^${_prefix}[0-9]+\\.[0-9]+\\.[0-9]+" "" VERSION_PRE_RELEASE "${TAG_SEGMENT_STR}")
+
+            if(VERSION_PRE_RELEASE MATCHES "^[^-]")
+                message(FATAL_ERROR "Invalid version string detected. ${VERSION_PRE_RELEASE} needs to start with a dash (${DESCRIBE})")
+            endif()
 
             if(DESCRIBE MATCHES "(-[0-9]+-g[0-9a-f]+)$")
               string(REGEX REPLACE "^.*-([0-9]+-g[0-9a-f]+)" "\\1" COMMIT_INFO "${DESCRIBE}")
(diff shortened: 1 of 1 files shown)


JonatanAntoni:
Thanks @Torbjorn-Svensson for raising this.
Can you please help me to understand the issue you try to resolve?

Glancing over the implementation the change looks weird to me. The `-` should not be optional according to SemVer rules.

If `VERSION_FULL` becomes `0.9.2-tools/packgen/0.9.2+p338-gd366fa6` this is a problem with specifying the release version tag prefix, properly. For a tool like packgen, the tag prefix must be set to `tools/packgen/`. This prefix should always get stripped from the version.

If for another tool the prefix is *not* set correctly, the latest tag is picked up regardless to which tool the tag belongs. 

A version tag for e.g. packgen must always use the prefix followed by a SemVer version:
- `tools/packgen/0.9.2` for a official patch release. This shall create version numbers like `0.9.2` if the tagged commit is used for the build. Or `0.9.2+p<N>-g<SHA>` if a later commit is used where <N> is the number of commits added since the tag and <SHA> is the commit id.
- tools/packgen/1.0.0-rc for a release candidate. This shall create version numbers like `1.0.0-rc0` if the tagged commit is used for the build. Or `1.0.0-rc<N>` if a later commit is used where <N> is the number of commits added since the tag.

While it looks like the proposed fix solves your immediate issue, I think the root cause for seeing a weird `FULL_VERSION` string is different.

Cheers,
Jonatan

JonatanAntoni:
I agree that it creates a wrong version string and we need to fix that.
My point here is the condition in https://github.com/Open-CMSIS-Pack/devtools/pull/994/files#diff-692d0df1d71c52c737c758a758bcc7da906e2258a0ff3ab2eea1dd976a3db46aL39-R41 should only match tags *with* a `-qualifier` part. I don't understand why a version tag that apparently does *not* have a `-qualifier` part every hits your changed condition. Let me try to reproduce.

Torbjorn-Svensson:
> I agree that it creates a wrong version string and we need to fix that. My point here is the condition in https://github.com/Open-CMSIS-Pack/devtools/pull/994/files#diff-692d0df1d71c52c737c758a758bcc7da906e2258a0ff3ab2eea1dd976a3db46aL39-R41 should only match tags _with_ a `-qualifier` part. I don't understand why a version tag that apparently does _not_ have a `-qualifier` part every hits your changed condition. Let me try to reproduce.

I think the reason is simply that the mandatory part `(-[a-zA-Z0-9-\\.]+)` of `(-[a-zA-Z0-9-\\.]+)(-[0-9]+-g[0-9a-f]+)?$` matches `-gd366fa6` in `tools/packgen/0.9.2+p338-gd366fa6`.

JonatanAntoni:
> I think the reason is simply that the mandatory part `(-[a-zA-Z0-9-\\.]+)` of `(-[a-zA-Z0-9-\\.]+)(-[0-9]+-g[0-9a-f]+)?$` matches `-gd366fa6` in `tools/packgen/0.9.2+p338-gd366fa6`.

Yes, indeed. The handling of versions with pre-release qualifier was incorrect. I fixed that in #997. Can you confirm?

Torbjorn-Svensson:
> > I think the reason is simply that the mandatory part `(-[a-zA-Z0-9-\\.]+)` of `(-[a-zA-Z0-9-\\.]+)(-[0-9]+-g[0-9a-f]+)?$` matches `-gd366fa6` in `tools/packgen/0.9.2+p338-gd366fa6`.
> 
> Yes, indeed. The handling of versions with pre-release qualifier was incorrect. I fixed that in #997. Can you confirm?

Yes, I confirm that it works.

Closing in favor of #997.

PR #1007, opened 2023-06-22
Define $generator_id and $pack_id variables in generator arguments

Torbjorn-Svensson:
$generator_id resolves to the current generator id
$pack_id resolves to the current pack id in YAML format

Also, add from-pack property on the generator nodes in the cbuild.yml to identify what pack the generator is defined in. This resolves potential conflicts when using generators from multiple packs that has the same name.

### DIFF
--- a/libs/rtemodel/src/RteGenerator.cpp
+++ b/libs/rtemodel/src/RteGenerator.cpp
@@ -145,12 +145,18 @@ string RteGenerator::GetExecutable(RteTarget* target, const std::string& hostTyp
 vector<pair<string, string> > RteGenerator::GetExpandedArguments(RteTarget* target, const string& hostType) const
 {
   vector<pair<string, string> > args;
+  string packId = GetPackage()->GetPackYamlID();
+  string generatorId = GetGeneratorName();
+
   RteItem* argsItem = GetArgumentsItem("exe");
   if (argsItem) {
     for (auto arg : argsItem->GetChildren()) {
       if (arg->GetTag() != "argument" || !arg->MatchesHost(hostType))
         continue;
-      args.push_back({arg->GetAttribute("switch"), ExpandString(arg->GetText())});
+      string res = arg->GetText();
+      RteUtils::ReplaceAll(res, "$pack_id", packId);
+      RteUtils::ReplaceAll(res, "$generator_id", generatorId);
+      args.push_back({arg->GetAttribute("switch"), ExpandString(res)});
     }
   }
   return args;
--- a/libs/rtemodel/src/RtePackage.cpp
+++ b/libs/rtemodel/src/RtePackage.cpp
@@ -743,6 +743,26 @@ string RtePackage::ConstructID()
 }
 
 
+string RtePackage::GetPackYamlID() const
+{
+  string res;
+
+  const auto& vendor = GetVendorString().empty() ? "" : GetVendorString() + "::";
+  const vector<pair<const char*, const string&>> elements = {
+    {"",  vendor},
+    {"",  GetName()},
+    {"@", GetVersionString()},
+  };
+
+  for (const auto& element : elements) {
+    if (!element.second.empty()) {
+      res += element.first + element.second;
+    }
+  }
+  return res;
+}
+
+
 string RtePackage::GetPackageID(bool withVersion) const
 {
   if (withVersion)
--- a/tools/projmgr/test/data/TestGenerator/ref/multiple-components.Debug+CM0.cbuild.yml
+++ b/tools/projmgr/test/data/TestGenerator/ref/multiple-components.Debug+CM0.cbuild.yml
@@ -30,6 +30,7 @@ build:
           category: sourceC
       generator:
         id: RteTestGeneratorIdentifier
+        from-pack: ARM::RteTestGenerator@0.1.0
         files:
           - file: ${CMSIS_PACK_ROOT}/ARM/RteTestGenerator/0.1.0/Templates/RteTestGen.c.template
             category: genSource
@@ -46,6 +47,7 @@ build:
       selected-by: RteTest:CORE
   generators:
     - generator: RteTestGeneratorIdentifier
+      from-pack: ARM::RteTestGenerator@0.1.0
       path: ../data/TestGenerator/MultipleComponents/RteTestGeneratorIdentifier
       gpdsc: ../data/TestGenerator/MultipleComponents/RteTestGeneratorIdentifier/RteTestGen_ARMCM0/RteTest.gpdsc
       command:
--- a/libs/rtemodel/include/RtePackage.h
+++ b/libs/rtemodel/include/RtePackage.h
@@ -448,6 +448,12 @@ public:
   void SetPackageState(PackageState packState) { m_packState = packState; }
 
   /**
+   * @brief get full pack ID in YAML format
+   * @return full pack ID in YAML format
+   */
+  std::string GetPackYamlID() const;
+
+  /**
    * @brief get full or common pack ID
    * @param withVersion flag to return full (true) or common (false) ID
    * @return full or common pack ID depending on argument
--- a/tools/projmgr/schemas/common.schema.json
+++ b/tools/projmgr/schemas/common.schema.json
@@ -478,12 +478,13 @@
       "description": "General generator info",
       "type": "object",
       "properties": {
+        "from-pack": { "$ref": "#/definitions/PackID"},
         "generator": { "type": "string", "description": "Section for a specific generator" },
         "path":      { "type": "string", "description": "Path name for storing the files generated" },
         "gpdsc":     { "type": "string", "description": "File name of the *.GDPSC file that stores the information in path" },
         "command":   { "$ref": "#/definitions/CommandType" }
       },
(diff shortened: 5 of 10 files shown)


jkrech:
In previous discussions it was our goal to make all information required by the generator via the cbuild.yml file of the current context. Why do we need these variables to be passed to the generator via the command line? What has changed and what current problem does it solve?

Torbjorn-Svensson:
> In previous discussions it was our goal to make all information required by the generator via the cbuild.yml file of the current context. Why do we need these variables to be passed to the generator via the command line? What has changed and what current problem does it solve?

In order for the generator application to collect the required info from the cbuild.yml, it needs to be able to properly identify what node to use. To avoid the ambiguity, the `$pack_id` symbol is used to identify in what pack the generator is defined (more than one pack could have a generator with the same name...).
The `$generator_id` is just for convenience to now have to write the same string over and over in the .pdsc file.

@slhultgren, please fill in if I missed something.

libs/rtemodel/src/RtePackage.cpp
```diff
@@ -743,6 +743,26 @@ string RtePackage::ConstructID()
 }
 
 
+string RtePackage::GetPackYamlID() const
```
brondani, line 746: Couldn't `string ProjMgrUtils::GetPackageID(const RteItem* pack)` be used instead of introducing this function?

libs/rtemodel/src/RteGenerator.cpp
```diff
@@ -145,12 +145,18 @@ string RteGenerator::GetExecutable(RteTarget* target, const std::string& hostTyp
...
+
   RteItem* argsItem = GetArgumentsItem("exe");
   if (argsItem) {
     for (auto arg : argsItem->GetChildren()) {
       if (arg->GetTag() != "argument" || !arg->MatchesHost(hostType))
         continue;
-      args.push_back({arg->GetAttribute("switch"), ExpandString(arg->GetText())});
+      string res = arg->GetText();
+      RteUtils::ReplaceAll(res, "$pack_id", packId);
+      RteUtils::ReplaceAll(res, "$generator_id", generatorId);
```
brondani, line 158: It seems `$pack_id` and `$generator_id` are deviating from the standard way of declaring/expanding/using variables such as `$P #P $S $D $B $G`:
https://open-cmsis-pack.github.io/Open-CMSIS-Pack-Spec/main/html/pdsc_generators_pg.html#element_generators
Concerning implementation look at [`RteCallback::ExpandString`](https://github.com/Open-CMSIS-Pack/devtools/blob/main/libs/rtemodel/src/RteCallback.cpp#L67).

@slhultgren We have discussed about the use of `[Gvendor][::Gtool][@Gversion]` as generator id. As long as the vendor guarantees the `Gtool` uniqueness, wouldn't it be sufficient?
https://github.com/Open-CMSIS-Pack/devtools/issues/434#issuecomment-1337608400

tools/projmgr/test/data/TestGenerator/ref/multiple-components.Debug+CM0.cbuild.yml
```diff
@@ -30,6 +30,7 @@ build:
           category: sourceC
       generator:
         id: RteTestGeneratorIdentifier
+        from-pack: ARM::RteTestGenerator@0.1.0
```
brondani, line 33: Isn't it redundant information already expressed in line 26?

libs/rtemodel/src/RtePackage.cpp
brondani: Not sure if we want to introduce the YAML format awareness into the `RtePackage` or if it's better to keep it limited to the `projmgr` tool.
As previously commented the variables expansion should be done via callbacks which can be tool's specific.
@edriouk can you please advise?

edriouk: We are already considering to unify component, API, and pack IDs across tools and libraries. RTE model will use then format used in .cproject/.csolution.   This happens not for 2.0 release, but soon. therefore I agree with @brondani  not to introduce extra methods in RTE model.

Torbjorn-Svensson: Do you have a way to fetch the generator of interest from the callback class?
As far as I can tell, only global context is available and no way to get the _current_ generator.

libs/rtemodel/src/RteGenerator.cpp
slhultgren: Hi @brondani,

I'm not sure to catch the point about Gvendor/Gtool/Gversion, since my problem currently with these are:
1. I don't understand what/how they are used
1.1. Gtool is defined as "Plain text name", this sounds almost like a human-readable description, not something to rely on functionally
1.2. There is no implementation yet supporting this AFAIK
2. They are still not guaranteed to be unique globally, so we would still require the pack id to make it fully unique
3. If Gvendor must follow the pack vendor, not sure why it is even possible to define it for the generator

Technically I guess Gvendor/Gtool/Gversion could work, as long as:
1. Open-CMSIS-Pack project guarantees that Only a single vendor can provide generators for a Gvendor
2. The Vendor guarantees that it does not release multiple conflicting generators using the same Gtool

but honestly I don't think any of these two points are feasible in practice (not worth the effort; The alternative of just allowing to specify the pack id is a lot simpler to maintain and double check on an organizational level. Each pack is more self-contained).

Currently my view on the Gvendor/Gtool/Gversion is that they provide some abstraction for calling generators (potentially from other packs??) but I don't think this level of abstraction is necessary or maybe not even useful?
Most likely when defining a component using a generator, we will know/expect some kind of behavior of the generator that we call, so most likely we just want to be able to call a generator from (at most) another pack (that we already know).

brondani: @slhultgren 
>1.1. Gtool is defined as "Plain text name", this sounds almost like a human-readable description, not something to rely on functionally

Not really different from other identifiers such as the pack name itself.

> 1.2. There is no implementation yet supporting this AFAIK

Right, if we want to benefit from the spec we should start implementing it. I thought we agreed [here](https://github.com/Open-CMSIS-Pack/devtools/issues/434#issuecomment-1337608400) this was the best approach to follow?

> 2.	They are still not guaranteed to be unique globally, so we would still require the pack id to make it fully unique

`Gvendor` should be unique. Each vendor would be responsible for their own `Gtool` identifiers.

> 3.	If Gvendor must follow the pack vendor, not sure why it is even possible to define it for the generator

I guess for the same reason it’s possible to specify `Cvendor` for components: if omitted it can be inherited from pack vendor, but it does not have to.

If you now strongly believe the `from-pack` is a better approach, I would recommend to update the specification.

slhultgren:
@jkrech @brondani
Hi, just looked at the recording of the meeting from 2023-08-22 at ~47 minutes, I think this PR is actually pretty close and aligned with the discussed proposal for the "generator-pack" and "generator-id" additions to the cbuild.yml.

If I understood correctly the presentation, the proposed generator-pack, and generator-id are actually a kind of "current-generator-pack" and "current-generator-id" since each invocation of a generator would get its' own unique cbuild.yml created for it.

The "minor" adjustments needed on this PR to align with that proposal I think would be to
1. Remove the $pack_id and $generator_id variables from the .pdsc generator argument specification, no need anymore since effectively replaced by the "cbuild.yml - generator-pack/id" (and also close https://github.com/Open-CMSIS-Pack/Open-CMSIS-Pack-Spec/pull/238)
2. The `generator-pack` and `generator-id` properties should be added to the cbuild.yml when running a generator.
2.1. Proposal is that since the cbuild.yml could be created for contexts without any generators as well, perhaps it would make sense to group these two properties in an optional object, eg.
```yml
build:
  ...
  current-generator:
    from-pack: Keil::STM32H7xx_DFP@3.1.1
    id: STM32CubeMX 
```
Where the `current-generator` object is optional, but the `from-pack` and `id` properties on the object are both mandatory if the object is included.
(Naming to be discussed as always 😄)

Implications:
1. The cbuild.yml when running a generator is always `generator+context` specific and always temporary (similar to the general idea of the other PR with cbuild.resolved.yml #975)

jkrech:
What if we take this one step further by creating a cbuild.yml file dedicated to an invoked generator, located side by side with the cbuild.yml for the context named e.g.:

```
<context>.cbuild-<generator-id>.yml
```

which contains both the generatorID and the packID of the generator and provides absolute paths for all file references to ease the use by the generator. `$G` is replaced by a reference to the above file.

These generator specific *.cbuild-generatorID.yml files are considered `temporary` as they are most likely out-of-date after the generator has completed and csolution re-reads the context information.

Any thoughts?

jkrech:
We agreed during the technical project meeting to include the generatorID and packID of the generator into the `<context>`.cbuild-gen.yml file` instead of extending the list of $variables to be placed at the command line.
Instead the generator invocation will use `$G` to specify the location of the cbuild-gen.yml file for the given context.

adding e.g.:
```
build:
  generated-by: csolution version 2.1.0
  generator-pack: Keil::STM32H7xx_DFP@3.1.1
  generator-id: STM32CubeMX
```

Torbjorn-Svensson:
Superseded by #1129

PR #1063, opened 2023-07-12
Fix crash when same pack is located on 2 paths

Torbjorn-Svensson:
Scenario:
csolution.yml:
```yml
packs:
  - pack: ARM::CMSIS@5.9.0
    path: ./ARM.CMSIS.5.9.0/
```

cproject.yml:
```yml
packs:
  - pack: ARM.CMSIS@5.9.0
```

In the above scenario, the same pack is defined on 2 different paths. The one in the csolution.yml is a project local path, and for the cproject.yml, it can be found in the $CMSIS_PACK_ROOT.

As the RteModel does not allow "duplicates", the RtePackage pointer, for the second pdsc file to be loaded, will be released when calling RteModel->Validate() and thus, only instances that is known by the model should be contained in m_loadedPacks in the ProjMgrWorker instance.

### DIFF
--- a/tools/projmgr/src/ProjMgrKernel.cpp
+++ b/tools/projmgr/src/ProjMgrKernel.cpp
@@ -127,6 +127,11 @@ bool ProjMgrKernel::LoadAndInsertPacks(std::list<RtePackage*>& packs, std::list<
   }
 
   globalModel->InsertPacks(newPacks);
-  packs.insert(packs.end(), newPacks.begin(), newPacks.end());
+
+  // Track only packs that were actually inserted into the model
+  packs.clear();
+  for (const auto& [_, pack] : globalModel->GetPackages()) {
+    packs.push_back(pack);
+  }
   return true;
 }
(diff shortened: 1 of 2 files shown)


tools/projmgr/src/ProjMgrKernel.cpp
```diff
@@ -127,6 +127,15 @@ bool ProjMgrKernel::LoadAndInsertPacks(std::list<RtePackage*>& packs, std::list<
...
+
+  // Only append packs that were actually insterted into the model
+  for (const auto& pack : newPacks) {
+    const string& packId = pack->GetPackageID();
+    if (globalModel->GetPackage(packId) == pack) {
+      if (find(packs.cbegin(), packs.cend(), pack) == packs.cend()) {
+        packs.push_back(pack);
+      }
+    }
+  }
```
brondani, line 139: To ensure the proper model inserted packs are tracked as loaded packs, I would suggest to use the `GetPackages()` function:
```suggestion
  // Track only packs that were actually inserted into the model
  packs.clear();
  for (const auto& [_, pack] : globalModel->GetPackages()) {
    packs.push_back(pack);
  }
```

brondani (commented):
@Torbjorn-Svensson Thanks for this finding and contribution.
Please see my suggestion below and also add or extended a test case to cover the new code.

brondani (approved):
The code changes are fully covered:
https://app.codecov.io/gh/Open-CMSIS-Pack/devtools/pull/1063?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=Open-CMSIS-Pack

edriouk (approved):

PR #1068, opened 2023-07-18
Add missing includes

maribu:
Using integer types with specific width requirements requires such as `uint32_t` requires including `<cstdint>`.


JonatanAntoni (approved):

maribu:
Thx :)

PR #1090, opened 2023-08-17
Allow additional properties in project schemas

ErikMallbergSTM:
Having additionalProperties set to false in schemas for project files makes backwards/future compatibility a problem. For example, a tool written to handle project files according to version 1.6 will fail to validate files created by tools for the newer version 1.7 if the latter has added new properties to some objects. This can be seen as unnecessarily strict, since added properties rarely is a breaking change.

### DIFF
--- a/tools/projmgr/schemas/common.schema.json
+++ b/tools/projmgr/schemas/common.schema.json
@@ -174,7 +174,6 @@
         "base-name": { "type": "string", "description": "Base name for all output files" },
         "type": { "$ref": "#/definitions/ArtifactsType" }
       },
-      "additionalProperties": false,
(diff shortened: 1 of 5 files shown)


brondani:
Disallowing additional properties is important for YAML syntax highlighting in IDE editors.
It's an immediate feedback when a node does not exist, is incorrectly placed or mistyped.
I would suggest to bring the matter to a wider audience at the weekly discussions.

image

jkrech:
The tools should be able to enable/disable the schema check as required by the user.

ErikMallbergSTM:
Closing this. We will patch this internally when/if needed.

PR #1099, opened 2023-08-28
Update AC6.6.18.0.cmake

jkrech:
https://github.com/Open-CMSIS-Pack/cmsis-toolbox/issues/69 Add linker command for Arm Assembler only projects.


jkrech:
I appreciate that we have little interest in maintaining support for the deprecated Arm Assembler Syntax files but it looks like there is examples out there in packs that get converted to csolution.

brondani:
This is a valid workaround but the proper definitive solution to the issue https://github.com/Open-CMSIS-Pack/cmsis-toolbox/issues/69 should be located in the `AS_ARM` dialect files in [CMakeASM](https://github.com/Open-CMSIS-Pack/devtools/tree/main/tools/buildmgr/cbuildgen/config/CMakeASM) and a regression test case should be added.

brondani:
Closing in benefit of https://github.com/Open-CMSIS-Pack/devtools/pull/1101.

PR #1137, opened 2023-09-27
Make files with different number of lines unequal

ErikMallbergSTM:
Previously, files with different number of lines could be considered equal, since the comparison stopped with success when any of the files ended.


soumeh01 (approved):
LGTM

PR #1165, opened 2023-10-25
Update codeql-analysis.yml

soumeh01:
CodeQL should run on the PR branches


soumeh01:
This change is added in the already opened PR

Issue #1146, opened 2023-10-11
[buildmgr] IAR single precision FPU error


brondani:
Reported by @RobertRostohar.

When the Cortex-M55 device attribute `Dfpu` is set to `SP_FPU` the IAR build leads to an error due to this mapping introduced in https://github.com/Open-CMSIS-Pack/devtools/pull/798:
[`set(FPU_FLAGS    "none" "ERROR" "VFPv5_d16" "VFPv5_d16")`](https://github.com/Open-CMSIS-Pack/devtools/pull/798/files#diff-78f7e0854be7e1fa01e09e7df721f38981ea70822556ad8e521e402a15e91000R105)
It seems it works with `VFPv5_sp`.
@jlonnberg @mario-pi Can you please advise?
And what about Cortex-M85? Thanks

bastkerg:
Hello! I am currently looking into this from @IARSystems side. What behavior would be expected?
The hardware is double or nothing. 
an in our IDE/Debugger that's what we support. 

If VFPv5_sp works in the compiler would be OK, but that support could break since the rest of our software don't support it right now.

If we would point it to VFPv5_d16 would that break someone's use case or be unexpected?

bastkerg:
See pull request:
Since hardware is double or nothing:

- Selecting SP_FPU should map onto DP hardware on Cortex-M55 and M85.
- Did not change the behavior for A-series processors. Here ARMCLANG maps SP_FPU -> NO_FPU which hardly seems correct, so throwing error here does not seem too bad.
- R52 and R52+ Does not support SP_FPU yet, it's a known issue EWARM-9425

brondani:
Fixed in https://github.com/Open-CMSIS-Pack/devtools/pull/1180

PR #1180, opened 2023-10-30
IAR single precision FPU error

bastkerg:
#1146 
Since hardware is double or nothing:
- Selecting SP_FPU should map onto DP hardware on Cortex-M55 and M85. 
- Did not change the behavior for A-series processors. Here ARMCLANG maps SP_FPU -> NO_FPU which hardly seems correct, so throwing error here does not seem too bad.
- R52 and R52+ Does not support SP_FPU yet, it's a known issue EWARM-9425


brondani (approved):
LGTM
@bastkerg Thanks for the contribution

Issue #1173, opened 2023-10-26
[csolution, cbuild] Introduce `--context-set` flag for backward compatibility


brondani:
The introduction of `*.cbuild-set.yml` files in [#1127](https://github.com/Open-CMSIS-Pack/devtools/issues/1127) can potentially lead to unexpected behavior.
To avoid breaking existing workflows the use of a flag `-S, --context-set` for enabling the `cbuild-set` feature is suggested.
In conclusion: the mechanism described in [#1127](https://github.com/Open-CMSIS-Pack/devtools/issues/1127), whether for reading, writing or updating should be enabled only when the `-S` or `--context-set` flag is present in the command line.

PR #1188, opened 2023-11-07
[projmgr] Implemented -S,--context-set option

grasci-arm:
https://github.com/Open-CMSIS-Pack/devtools/issues/1173

* [projmgr] Add -S,--context-set option

* Addressed review comments


brondani (approved):
LGTM

PR #1225, opened 2023-11-24
[projmgr] Add --frozen-packs argument to csolution

Torbjorn-Svensson:
The `--frozen-packs` option is only applicable for the following commands:
- `convert`
- `update-rte`

If `cbuild-pack.yml` needs to be created or updated, then `--frozen-pack` will treat this as an error and fail the command.

### DIFF
--- a/libs/rtemodel/src/RteProject.cpp
+++ b/libs/rtemodel/src/RteProject.cpp
@@ -436,14 +436,18 @@ void RteProject::AddCprjComponents(const Collection<RteItem*>& selItems, RteTarg
   // update file versions
   for (auto [instanceName, fi] : m_files) {
     string version;
+    auto rteFile = fi->GetFile(target->GetName());
     auto itver = configFileVersions.find(instanceName); // file in cprj can be specified by its instance name
     if (itver == configFileVersions.end())
       itver = configFileVersions.find(fi->GetName()); // or by original name
     if (itver != configFileVersions.end()) {
       version = itver->second;
+    } else {
+      // Fall back to the version noted in the pack
+      version = rteFile->GetVersionString();
     }
     UpdateFileInstanceVersion(fi, version);
-    UpdateConfigFileBackups(fi, fi->GetFile(target->GetName()));
+    UpdateConfigFileBackups(fi, rteFile);
   }
 }
 
--- a/tools/projmgr/src/ProjMgrYamlEmitter.cpp
+++ b/tools/projmgr/src/ProjMgrYamlEmitter.cpp
@@ -35,7 +35,7 @@ protected:
   const string FormatPath(const string& original, const string& directory);
   bool CompareFile(const string& filename, const YAML::Node& rootNode);
   bool CompareNodes(const YAML::Node& lhs, const YAML::Node& rhs);
-  bool WriteFile(YAML::Node& rootNode, const std::string& filename);
+  bool WriteFile(YAML::Node& rootNode, const std::string& filename, bool allowUpdate = true);
   const bool m_useAbsolutePaths;
 };
 
@@ -705,7 +705,7 @@ bool ProjMgrYamlBase::CompareNodes(const YAML::Node& lhs, const YAML::Node& rhs)
   return (lhsData == rhsData) ? true : false;
 }
 
-bool ProjMgrYamlBase::WriteFile(YAML::Node& rootNode, const std::string& filename) {
+bool ProjMgrYamlBase::WriteFile(YAML::Node& rootNode, const std::string& filename, bool allowUpdate) {
   // Compare yaml contents
   if (RteFsUtils::IsDirectory(filename)) {
     ProjMgrLogger::Error(filename, "file cannot be written");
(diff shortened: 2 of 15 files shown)


Torbjorn-Svensson:
> @Torbjorn-Svensson, is this PR ready for review or in Draft? Do you think this should/must be included in CMSIS-Toolbox 2.2.0?

As we discusses on the meetings the previous 2 weeks, this was the followup PR that should preferably be included in the 2.2.0 release, but was "optional" in case time did not allow to make it stable.
The only thing that is in "draft" mode of this PR is the naming. I hope that there will be some feedback on the PR if the name `--frozen-packs` is acceptable or if there is something else that would fit the need better.

jkrech:
Alternative names for the option I can think of are:

1. 

```
--no-update-packs
```
similar to rte file update:
```
--no-update-rte
```
2.  
```
--fixed-packs
```

tools/projmgr/src/ProjMgrYamlEmitter.cpp
```diff
@@ -721,6 +721,10 @@ bool ProjMgrYamlBase::WriteFile(YAML::Node& rootNode, const std::string& filenam
     }
   }
   else if (!CompareFile(filename, rootNode)) {
+    if (!allowUpdate) {
+      ProjMgrLogger::Error(filename, "file not allowed to be updated");
```
brondani, line 725: Would it be possible to be more specific about the reasons why it cannot be updated (e.g. which pack/version is required vs updated) maybe in `verbose` mode?

Torbjorn-Svensson: Is this required in this PR? Both `yarn` and `cargo` does not specify the difference.

brondani: It is not required, it is just a suggestion, what is your opinion? Regardless of what `yarn` and `cargo` do, would it bring helpful information to the user?

Torbjorn-Svensson: I suppose it might indicate why the CI build failed, but the user would still need to do the same step locally in order to update the `*.cbuild-pack.yml` file.
Can you please create an issue for this, as I believe it should not be part of this PR.

brondani: Just for clarity my comment concerns the changes in lines 724-725 of [`ProjMgrYamlEmitter.cpp`](https://github.com/Open-CMSIS-Pack/devtools/pull/1225/files/e5eced6777ff4602b9a4b21a5b27a73403e8c868#diff-622f5ce2a380acef913dfa78872551356f6b1be48e286f33de05da865078e738R724-R725):
```
    if (!allowUpdate) {
      ProjMgrLogger::Error(filename, "file not allowed to be updated");
```
In other words I am asking whether it is meaningful to increase the verbosity of this error message that is being introduced in this PR so the question is specific to this PR.
Also it should be indifferent whether it runs on CI or locally.

edriouk:
This PR can make use of "Dedicated method to Update RTE folder" #1244

jkrech:
In the technical meeting from 2023-12-05 we agreed to use the command line option:

`--frozen-packs`

Torbjorn-Svensson:
@edriouk Can you please take a look at the 2nd commit in this PR as it appears to fix the issue that I talked about in the meeting today.

edriouk (commented):
@Torbjorn-Svensson, I need more information about the change in RteProject.cpp. 
To my understanding, --frozen-packs argument is used against already converted projects all config files are already copied to the RTE directory and all @base files are created. 
Could you explain the case you are trying to fix? 
@brondani, could you please approve the rest.

libs/rtemodel/src/RteProject.cpp
brondani, line 1: Please take a look at the indirect changes caught by `codecov` in CbuildModel.cpp:
https://app.codecov.io/gh/Open-CMSIS-Pack/devtools/pull/1225/indirect-changes

It seems it has been incorrectly covered by test cases such as the [GCC/TrustZone/CM33_s](https://github.com/Open-CMSIS-Pack/devtools/tree/main/tools/buildmgr/test/testinput/Examples/GCC/TrustZone/CM33_s) and with the changes here it has arisen.
I would recommend to add or extend a cbuildgen test case to cover config files that have new versions. For example adding an older base version `gcc_arm.ld.base@1.0.0` to the mentioned test case would cover it.

libs/rtemodel/src/RteProject.cpp
```diff
@@ -436,14 +436,18 @@ void RteProject::AddCprjComponents(const Collection<RteItem*>& selItems, RteTarg
...
     string version;
+    auto rteFile = fi->GetFile(target->GetName());
     auto itver = configFileVersions.find(instanceName); // file in cprj can be specified by its instance name
     if (itver == configFileVersions.end())
       itver = configFileVersions.find(fi->GetName()); // or by original name
     if (itver != configFileVersions.end()) {
       version = itver->second;
+    } else {
+      // Fall back to the version noted in the pack
+      version = rteFile->GetVersionString();
```
Torbjorn-Svensson, line 447: @edriouk You can check out this commit and give the unit-tests a try with out this line and you will see a failure in the test case `RunProjMgr_UpdateRte`.
The wrong base version would be selected if the RTE directory tree does not yet exist.

libs/rtemodel/src/RteProject.cpp
Torbjorn-Svensson: I think my last push should solve this.

Torbjorn-Svensson:
Rebased on projmgr/2.2.1. There should be no more changes needed before this PR can be merged.
Please have a look at this and let me know if there is anything more I need to fix.

brondani (approved):
@Torbjorn-Svensson Thanks for the contribution.
LGTM

PR #1227, opened 2023-11-27
[projmgr] Update default linker scripts

MiloradCvjetkovic:
Add or update support for uninitialized region


JonatanAntoni (approved):
Looks good to me by glancing over.

PR #1231, opened 2023-11-29
[projmgr] Update default linker script for AC6

MiloradCvjetkovic:
- added .bss at beginning of the .noinit section names as required by Arm Compiler 6


brondani (approved):
When the section has the name `.noinit` its type is set as "Data" while when placing it in `.bss.noinit` the type is "Zero".

PR #1257, opened 2024-01-04
[buildmgr] Update CMSIS version requirement for AC6 examples (#762)

grasci-arm:


brondani (approved):

jeromecoutant:
Examples doesn't support CMSIS 6.0 ?

brondani:
> Examples doesn't support CMSIS 6.0 ?

Yes, these are old CPRJ examples that will be deprecated together with CPRJ handling but should maintained for the time being.

PR #1283, opened 2024-01-17
[rtemodel] Support running generators when there is a space in the path

slhultgren:
Support running generators when there is a space in the path to the generator executable.
Renamed the folder of the generator tests to contain spaces.


edriouk (approved):
Looks good, thank you for the contribution

PR #1296, opened 2024-01-26
Fix nightly build artifact tags

grasci-arm:


JonatanAntoni (dismissed):

soumeh01 (changes requested):
Please don't merge this as there is one more commit change required to be made

Issue #1291, opened 2024-01-23
[csolution] Clarify precedence of controls for pack version selection


jkrech:
**Describe The Problem To Be Solved**
There is a number of controls for selecting pack versions loaded by `csolution`.
a) In the CMSIS solution project files (*.csolution.yml, *.cproject.yml, *.clayer.yml) under the `packs:` node
   - no `packs:` node
   - packs without version
   - packs with fixed version
   - packs with version range
b) Pack versions listed in *.cbuild-pack.yml (lock file)
c) `csolution` command line option `-l` / `--load`
   - all
   - latest
   - required

We need to document the precedence and expected outcomes when these are combined, especially when `-all` or `-latest` is defined.

**(Optional): Suggest A Solution**
A concise description of the expected results.

In addition it might be useful to have an option to print the list of the effectively listed pack versions.
a) as "debug" or "verbose" output of csolution commands
b) `csolution list packs <solution>.csolution.yml --load ...`

brondani:
Concerning the `--load` option:
- `--load all`: load all versions of all packs in the pack repository :warning: 
- `--load latest`: load the latest version of each pack in the pack repository and the packs listed in `packs:` node :warning: 
- `--load required`: load only packs listed in the `packs:` node, return error if empty

- no `--load` option and no `packs:` node: load the latest version of each pack in the pack repository (equivalent to `--load latest`) - now it's potentially impacted by *.cbuild-pack.yml.

when `packs:` node is used:
- packs without version: load latest installed version
- packs with fixed version: load the fixed version if installed
- packs with version range: load latest installed version in the version range

:warning: The command `csolution list packs <solution>.csolution.yml --load ...` should already list the packs that are effectively loaded, however there is a defect: in both cases `--load all` and `--load latest` all packs are loaded but after filtering only the latest remain. It appears correct in `list packs` but affects other commands. A bugfix is being proposed #TBD.

@Torbjorn-Svensson can you please comment about the impact of *.cbuild-pack.yml when loading packs?

ReinhardKeil:
Current proposal is:
  - when option `--load` is used the `*.cbuild-pack.yml` is not read, but is updated when the covert is error free.

ReinhardKeil:
Should we create a file `*.cbuild-pack.yml.bak` when a new `*.cbuild-pack.yml` is created with updated versions.  It might be not required due to #1306

PR #1302, opened 2024-01-30
[projmgr] Don't consider cbuild-pack.yml with --load <all|latest> (#1291)

Torbjorn-Svensson:
When `--load all` or `--load latest` argument is provided, the content of `cbuild-pack.yml` files should not be considered, but the file should be updated with the new resolved packs.


brondani (approved):
@Torbjorn-Svensson Thanks for the contribution.
LGTM

PR #1378, opened 2024-03-18
Fix typos in iar_linker_script.icf.src

joseph-yiu:
ROM1 and RAM1 have incorrect BASE (used ROM0 and RAM0 instead. Corrected to use ROM1 and RAM1 base.


jkrech (approved):
Looks correct/consistent now.

Issue #1339, opened 2024-02-27
Pack version does not include meta-data in `cbuild*.yml` files


jeromecoutant:
Hi

I got an issue during the convert procedure with packs with a version with metadata

csolution.yml file:
`    - pack: STMicroelectronics::mw_mbed_crypto`

$ grep mw_mbed_crypto $CMSIS_PACK_ROOT/.Local/local_repository.pidx
`    <pdsc vendor="STMicroelectronics" name="mw_mbed_crypto" version="2.28.4+st.0.1.1" url="file:///C:/xxx/"/>`

$ csolution convert -s xxx.csolution.yml -c xxx.GCC+B-U585I-IOT02A

generated cprj file has correct information:
    `<package name="mw_mbed_crypto" vendor="STMicroelectronics" version="2.28.4+st.0.1.1:2.28.4+st.0.1.1"/>`

But generated cbuild-pack.yml is wrong:
`    - resolved-pack: STMicroelectronics::mw_mbed_crypto@2.28.4`

And generated cbuild.yml as well:
`    - pack: STMicroelectronics::mw_mbed_crypto@2.28.4`

- [tool or library]
csolution 2.2.1 for Windows

edriouk:
According to Semantic Versioning meta data should be ignored by any comparison making the following versions equal 
`6.0.0  == 6.0.0+meta1 ==  6.0.0+meta2`
Since Pack ID is constructed as `Vendor::Name@Version`, meta data should be stripped from the version.
The  cprj file stores pack attributes taken from the pdsc, Pack ID  gets constructed out of that later with meta-data removed.
In YAML files we always store already constructed Pack ID => it may not contain meta data.

It is possible to obtain meta-data from corresponding pack object by calling` RteItem::GetVerison() `method.
However, storing the meta data as a part of ID in YAML files  may lead to forward compatibility issues, because not all tools correctly process it, e.g. [[cpackget] must strip meta data from version ](https://github.com/Open-CMSIS-Pack/cpackget/issues/268)

Torbjorn-Svensson:
Hi @edriouk,

According to https://semver.org, it's defining that build metadata should not be used for precedence, however, equality is equality and not the same as precedence.

So, from understanding is that we need to store also the metadata in the cbuild*.yml files to ensure that the same pack will be selected next time when using the cbuild-pack.yml file.

edriouk:
[...]

PR #1382, opened 2024-03-18
[projmgr] Strip build metadata when handling pack versions

grasci-arm:
Strip build metadata when handling pack versions.
Throw a warning when specified metadata does not match the one from the loaded pack.
Addresses https://github.com/Open-CMSIS-Pack/devtools/issues/1339#issuecomment-1987962578


jkrech (approved):
LGTM -> no meta information in cbuild*.yml files other than selected-by and warning if explicitly set meta-data does not match the meta data of the loaded pdsc file.

PR #1408, opened 2024-03-28
cbuild-idx listing 'errors' & 'packs-missing' status report

grasci-arm:


tools/projmgr/include/ProjMgrYamlEmitter.h
```diff
@@ -32,7 +32,8 @@ class ProjMgrYamlEmitter {
    * @return true if executed successfully
   */
   static bool GenerateCbuildIndex(ProjMgrParser& parser,
-    const std::vector<ContextItem*> contexts, const std::string& outputDir);
+    const std::vector<ContextItem*> contexts, const std::string& outputDir,
```
JonatanAntoni, line 35: Is passing `std::verctor<> contexts` by value intentional? If yes, the `const` doesn't have any effect. I'd rather expect passing by reference to avoid copying.

```suggestion
    const std::vector<ContextItem*>& contexts, const std::string& outputDir,
```

tools/projmgr/src/ProjMgrYamlEmitter.cpp
```diff
@@ -895,20 +897,21 @@ bool ProjMgrYamlBase::WriteFile(YAML::Node& rootNode, const std::string& filenam
 }
 
 bool ProjMgrYamlEmitter::GenerateCbuildIndex(ProjMgrParser& parser,
-  const vector<ContextItem*> contexts, const string& outputDir) {
+  const vector<ContextItem*> contexts, const string& outputDir,
```
JonatanAntoni, line 900: Same question here. Passing by reference to avoid copying the vector? Or remove the `const` to express copy is intended.

```suggestion
  const vector<ContextItem*>& contexts, const string& outputDir,
```

JonatanAntoni (commented):
Please double check function parameters. This might be a separate PR because my comments don't apply to changed code.

edriouk (approved):

Issue #1356, opened 2024-03-04
unclear warning csolution: required pack 'ARM::CMSIS-Compiler@2.0.0' is not loaded


ReinhardKeil:
`cbuild` creates warnings `required pack is not loaded`. IMHO these warning is not helpful and it should be consider to remove it.

```
cbuild .\TestM85\TestM85.csolution.yml --context-set --update-rte --packs --schema --context Project.Debug+SSE-310-MPS3_FVP
info cbuild: Build Invocation 2.2.1 (C) 2023 Arm Ltd. and Contributors
warning csolution: context 'Project.Debug+SSE-310-MPS3_FVP': required pack 'ARM::CMSIS-Compiler@2.0.0' is not loaded
warning csolution: context 'Project.Debug+SSE-310-MPS3_FVP': required pack 'ARM::DMA350@1.0.0' is not loaded
```

**To Reproduce**
Use [TestM85.zip](https://github.com/Open-CMSIS-Pack/devtools/files/14484021/TestM85.zip) and run `cbuild` as shown above.

**Environment:**

- CMSIS-Toolbox 2.2.1
- OS: Windows

jkrech:
The background for the above warnings is, that the `ARM.V2M_MPS3_SSE_310_BSP.pdsc` of version 1.4.0 specifies these pack requirements:

``` xml
  <requirements>
    <packages>
      <package vendor="ARM" name="CMSIS" version="6.0.0" />
      <package vendor="ARM" name="CMSIS-Compiler" version="2.0.0" />
      <package vendor="ARM" name="DMA350" version="1.0.0"/>
    </packages>
  </requirements>
```

I think the message needs revising to:

```
warning csolution: context 'Project.Debug+SSE-310-MPS3_FVP': the pack 'ARM::CMSIS-Compiler@2.0.0' required by pack 'ARM.V2M_MPS3_SSE_310_BSP@1.4.0' is not specified
```

and maybe `csolution` should only print it if a component cannot be resolved or a component dependency cannot be resolved. Basically this gives a hint to the user to add these packs to the YML Input file's `packs:` node.

edriouk:
Fixed by PR https://github.com/Open-CMSIS-Pack/devtools/pull/1364

ReinhardKeil:
I believe this error should not appear when no components from the required pack are used in the csolution.
To me this warning is misleading, as it forces uses to specify packs that are unused.

edriouk:
[...]

PR #1410, opened 2024-03-28
Fix for issue #1356 (#849)

grasci-arm:
unclear warning csolution: required pack 'ARM::CMSIS-Compiler@2.0.0' is not loaded #1356

The warnings are now printed only if -d (debug) switch is specified


jkrech (approved):
LGTM

PR #1448, opened 2024-04-17
Improved defines list calculation to only match key

grasci-arm:
Improved defines list calculation to only exact key match


brondani (approved):
LGTM

Issue #1442, opened 2024-04-16
`csolution convert` crashes when using attr=config files for generated component


Torbjorn-Svensson:
**Describe the bug**
When using a .pdsc with a file definition for a name that does not exist in the resulting .gpdsc file, the `csolution convert` call crashes as it's unable to identify the version of the file.

**To Reproduce**
- Have a .pdsc file with the following statement:
`<file category="sourceC" name="foo-config.c" version="1.0.0" attr="config"/>`
- Have a .gpdsc that does _not_ have any file definition for `foo-config.c`.
- Both .pdsc and .gpdsc have a component with the same name.

Run `csolution convert` on a project that includes the component defining the above file.

**Expected behavior**
The `csolution convert` should not crash the application and the `foo-config.c` from the .pdsc should not be created in RTE since it's not defined in .gpdsc?
What is expected behavior here?

**Environment (please complete the following information):**

- csolution
- Version 2.3.0 or later
- OS: windows and linux, darwin?

**Additional context**
None.

jkrech:
@Torbjorn-Svensson  thanks a lot for adding a reproducer for the crash. At the moment everyone is busy on other topics but it is understood that you do not plan to work on this issue any further.

Torbjorn-Svensson:
> The crash, however, should be fixed

It looks like the crash is still not fully fixed. See https://github.com/Open-CMSIS-Pack/devtools/pull/1461#issuecomment-2151923329

jkrech:
@iomint confirmed that the crash can no longer be reproduced, however there is additional tests to be executed before it is clear whether the use cases of using config files in gpdsc files is sufficiently covered.

PR #1461, opened 2024-04-23
[projmgr] Expose crash when running csolution convert (#1442)

Torbjorn-Svensson:


Torbjorn-Svensson:
@jkrech Here is the reproducer that was asked of me to create in the meeting last week.

test/packs/ARM/RteTestGenerator/0.1.0/ARM.RteTestGenerator.pdsc
```diff
@@ -187,6 +187,12 @@
     <component generator="RteTestExternalGenerator" Cclass="RteTestGenerator" Cgroup="Check Global Generator" Cversion="0.9.0">
       <description>Component addressing external</description>
     </component>
+    <component generator="RteTestGeneratorIdentifier" Cvendor="ARM" Cclass="Device" Cgroup="RteTest Generated Component" Csub="RteTestOverlap" Cversion="1.1.0">
+      <description>Component with overlaping file lists</description>
+      <files>
+        <file category="sourceC" name="Src/RteTestConfig.c" attr="config" version="1.1.0"/>
```
edriouk, line 193: attr="config" is not supported for component files in gpdsc: the files  are alsready in project workspace and should not be copied

edriouk (changes requested):

edriouk:
The tests are failing because a component has been added. This requires updates of golden data in some tests.
In particular, RteModelUtinTests.RteChkTest.Summary require changing number of components from 57 to 58

test/packs/ARM/RteTestGenerator/0.1.0/ARM.RteTestGenerator.pdsc
Torbjorn-Svensson: While that might be so, this is how I could easily make the crash visible.
If this part is removed, make sure that csolution still crashes on the same statement as without it to ensure that the reported crash is corrected.

Torbjorn-Svensson:
If the .pdsc contains the `attr="config"` on a file that is in a component that is replaced by the gpdsc, then `csolution convert` still crashes in current main (2275f4afe4bf73154dacd5ca4276f925b45e149a), so the crash is still not fixed!

@jkrech I would like to revert the removal of `attr="config"` you did on my branch to re-expose the crash that should be addressed. My PR was about exposing a crash, not "correcting" the wrong usage in the input files. WDYT?

jkrech:
I reverted the inappropriate change to expose the crash.

edriouk (approved):

PR #1465, opened 2024-04-24
[rtemodel] Align implementation with specification

Torbjorn-Svensson:
See https://github.com/Open-CMSIS-Pack/Open-CMSIS-Pack-Spec/issues/307

**Note**: This changes both the RtePackage.cpp file, but also the pdsc files used for testing, including something that looks like it was done to ensure that the releases were sorted.


edriouk (approved):
Looks good

Issue #1300, opened 2024-01-30
Clarify the difference between add-path and files with include category


mcgordonite:
**Describe The Problem To Be Solved**

The Toolbox documentation specifies two ways to add to the include path:
1. The [add-path](https://github.com/Open-CMSIS-Pack/cmsis-toolbox/blob/main/docs/YML-Input-Format.md#add-path) property.
2. A file with category "include": https://github.com/Open-CMSIS-Pack/devtools/blob/76d2052c3a5e5898e3cc17f0b61fadedf610e823/tools/projmgr/schemas/common.schema.json#L488-L491

The documentation could be clearer on which is the preferred method.

**(Optional): Suggest A Solution**

The file categories are in the schema but not documented in the user guide. Should they be removed from the schema?

jkrech:
The description of FileCategoryType mentions that it is used for component files and therefore are used in *.cbuild.yml files in the context of component node but not in csolution, cproject or clayer CMSIS solution project files.

mcgordonite:
> The description of FileCategoryType mentions that it is used for component files and therefore are used in *.cbuild.yml files in the context of component node but not in csolution, cproject or clayer CMSIS solution project files.

In the schema, `FileCategoryType` is used in `FileType`, which is eventually used by `GroupsType` to describe the user files in cproject or clayer files. Perhaps there's an error there, and there should be two different `FileType` definitions in the schema.

jkrech:
@mcgordonite is this a blocker for you? What is the impact? Does this need to be fixed in 2.3.0

mcgordonite:
This is not a blocker, but it is potentially redundancy in the spec that might confuse users. Not required for 2.3.0 if it won't fit.

brondani:
As per my [comment in #1471](https://github.com/Open-CMSIS-Pack/devtools/pull/1471#pullrequestreview-2038015475) the `include` and other `includeX` enums could/should be removed from the schema for input user `files`, since `add-path` is preferred for the same purpose.
However the removal of enums from the schema potentially breaks backward compatibility, for this reason I would suggest to reject the PR which anyway needs rework and postpone the change to the next [major increase](https://github.com/orgs/Open-CMSIS-Pack/projects/12/views/1).

jkrech:
I confirm that in this issue, we are only considering to remove `FileCategoryType` from files specified in the `<files>` section of the CMSIS Solution YAML input files: 
- `*.csolution.yml`
[...]

PR #1471, opened 2024-04-25
Remove include file category type

mcgordonite:
A potential solution to https://github.com/Open-CMSIS-Pack/devtools/issues/1300.

This PR removes `"include"` from the `FileCategoryType` enum in the schema. The documentation does not refer to this, so it can be argued this is not a breaking change.

Questions:
1. Should the other `"includeX"` values be removed too?
2. Is there any code to remove in the implementation of `csolution`?
3. Should we remove the `category` entirely, and rely on file extension instead? This is consistent with the documentation here: https://github.com/Open-CMSIS-Pack/cmsis-toolbox/blob/main/docs/YML-Input-Format.md#filename-extensions

### DIFF
--- a/tools/projmgr/schemas/common.schema.json
+++ b/tools/projmgr/schemas/common.schema.json
@@ -540,7 +540,7 @@
       "items": { "$ref": "#/definitions/FileType" }
     },
     "FileCategoryType": {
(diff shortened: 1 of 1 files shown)


brondani (changes requested):
It is not yet documented but the file `category` in input yml files is accepted and can be used to explicitly set the file's category when the [file extension](https://github.com/Open-CMSIS-Pack/cmsis-toolbox/blob/main/docs/YML-Input-Format.md#filename-extensions) is not among the supported extensions or when it should be handled differently. @ReinhardKeil Would you consider documenting it?

@mcgordonite The schema element `FileCategoryType` being changed in this PR, as well as its parents `FileType` and `FilesType` are not used only in input yml files. After this change it cannot be shared anymore and should be separated to suit the different cases.
For consistency also the other `includeX` should be removed.

Issue #1468, opened 2024-04-25
Regions header file not processing information from `<boards>`


RobertRostohar:
**Describe The Problem To Be Solved**
It seems that `<memory>` information from `<boards>` is not considered when creating the regions header file (only `<memory>` information from `<devices>` is considered).

**(Optional): Suggest A Solution**
`<memory>` information from `<boards>` should also be processed when `board` is specified as `target-type`.

edriouk:
@RobertRostohar , I need a reproducer. We have a test that adds board memory to the file. That is validated at every PR, no fails have been seen so far..

RobertRostohar:
I have seen this issue with B-U585I-IOT02A BSP. It should be easy reproducible.

You can use the following example [Hello_B-U585I-IOT02A](https://github.com/Arm-Examples/Hello_B-U585I-IOT02A):
- modify the BSP pack version to 1.1.0 (includes also memory definitions):
  [hello.csolution.yml](https://github.com/Arm-Examples/Hello_B-U585I-IOT02A/blob/main/hello.csolution.yml#L10): `- pack: Keil::B-U585I-IOT02A_BSP@1.1.0`
- enable automatic linker script generation by adding to [hello.cproject.yml](https://github.com/Arm-Examples/Hello_B-U585I-IOT02A/blob/main/hello.cproject.yml):
  ```
  linker:
    - auto:
  ```
  
After building, `RTE\Device\STM32U585AIIx\regions_B_U585I_IOT02A.h` contains only memory defines from the DFP. Memory `RAM-External` and `Flash-External` from BSP are not considered.

edriouk:
@RobertRostohar , where do I get Keil::B-U585I-IOT02A_BSP@1.1.0?
I see 1.0.0 as the latest version

jkrech:
The problem got fixed and will be released in CMSIS-Toolbox csolution 2.4.0.
(Tested with version 2.4.0-devint1)

Issue #888, opened 2023-05-05
[csolution, cbuildgen] Unicode characters are not supported


brondani:
**Describe the bug**
When trying to convert a project with Unicode characters `csolution` throws an exception probably related to [std::filesystem::path::generic_string()](https://developercommunity.visualstudio.com/t/stdfilesystempathgeneric-string-throws-an-exceptio/721120).
Trying to build a CPRJ with `cbuildgen` results in the following error:
```
*** ERROR M204:
  Path not found: 'C:/Temp/unicode/Unicode/技术/main.c'!
```

**To Reproduce**
Use the attached [Unicode.zip](https://github.com/Open-CMSIS-Pack/devtools/files/11404463/Unicode.zip)

**Environment**
- csolution v1.6.0, cbuildgen v1.6.0
- OS: windows

PR #1481, opened 2024-04-29
Resolve board in RteTarget (#888)

grasci-arm:
* Resolve board in RteTarget
fixes: Regions header file not processing information from <boards> #1468
Use consistent directory names with dots and dashes


jkrech (approved):
LGTM - got already reviewed internally.

Issue #1203, opened 2023-11-14
Proposal: restrict --context-set to one identical target-type and only one build-type per context.


ReinhardKeil:
**The Problem To Be Solved**

When working with multiple projects it is required to know which `--context` selection should be used to debug and download to the target hardware. For that reason we have introduced the option `--context-set`. See also #896

Currently it is possible to use more than one build-type per context. This makes the tool very flexible, but is hard to understand from user's point of view. IMHO, we should also consider how a debugger or a download process would use the generated projects.

**Suggested Solution**
When using the option `--context-set`, the CMSIS-Toolbox only allows one `build-type` per `context`.  For example https://github.com/Open-CMSIS-Pack/csolution-examples/tree/main/SimpleTrustZone that would mean it is impossible to request context `CM33_s.Release+AVH` and `CM33_s.Debug+AVH` at the same time.

For [Access Sequences](https://github.com/Open-CMSIS-Pack/cmsis-toolbox/blob/main/docs/YML-Input-Format.md#access-sequences) I propose:

- Access sequences that have no `build-type` qualifier are handled as follows:
  - without context-set: are only permitted when they refer to the same `project-name`. Otherwise an error or warning is generated
  - with context-set: are resolved to the selected build-type for the `project-name`, which means with context-set it is possible to mix build types.

See also: #1198

brondani:
My comment https://github.com/Open-CMSIS-Pack/devtools/issues/896#issuecomment-1651409667 is still valid.

mcgordonite:
Should we go further and only allow one target type and build type per project?

ReinhardKeil:
> Should we go further and only allow one target type and build type per project?

I guess that depends:
- IDE: Yes, perhaps in an IDE flow the right choice. Usually you have just one target connected. 
- DevOps/MLOps: CLI builds for CI/CD testing may need multiple targets

VGRSTM:
> Should we go further and only allow one target type and build type per project?

IDE view:  for target-type possibly one is good enough, build-type my guess is no choice we have to support multiples (common Debug & Release profiles are a must)

ReinhardKeil:
I believe this needs further review. We should make also a decision about `target-types` and maybe restrict them to just one when using `context-set`.

This is related to #1215

jkrech:
[...]

PR #1509, opened 2024-05-14
Restrict --context-set to one identical target-type and only one build-type per context

grasci-arm:
Addressing : https://github.com/Open-CMSIS-Pack/devtools/issues/1203

### DIFF
--- a/tools/projmgr/src/ProjMgr.cpp
+++ b/tools/projmgr/src/ProjMgr.cpp
@@ -541,6 +541,14 @@ bool ProjMgr::Configure() {
   if (!m_worker.ParseContextSelection(m_context, checkCbuildSet)) {
     return false;
   }
+
+  // validate and restrict the input contexts when used with -S option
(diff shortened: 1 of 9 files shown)


tools/projmgr/src/ProjMgr.cpp
```diff
@@ -530,6 +530,65 @@ bool ProjMgr::GenerateYMLConfigurationFiles() {
...
+bool ProjMgr::ValidateContexts(const std::vector<std::string>& contexts) {
+  if (contexts.empty()) {
+    return true;
+  }
+
+  std::string selectedTarget = "";
+  std::map<std::string, std::string> projectBuildTypeMap;
+  std::vector<std::string> errorContexts;
+
+  // Check 1: if all the contexts has same target type
```
jkrech, line 542: all contexts have ...

brondani (approved):

Issue #926, opened 2023-05-19
Create warning on potentially mis-spelled names in `for-context:`, `not-for-context:`, `for-compiler:`


ReinhardKeil:
This is a future improvement: when csolution compares  names in `for-context:`, `not-for-context:`, `for-compiler:` it should highlight when a name is potentially misspelled.

Example:
```yml
  target-types:
    - type: Board                          # When applying target-type: 'Board':
```

When using `for-context: +board` this will never match -> issue warning: `target-type name does not exist in solution`

When using `for-context: .Board2` this will never match -> issue warning: `build-type name does not exist in solution` (use obviously potentially mixed target with build type).

brondani:
Addressed in https://github.com/Open-CMSIS-Pack/devtools/pull/1121 and https://github.com/Open-CMSIS-Pack/devtools/pull/1126.

PR #1527, opened 2024-05-21
Do not create RTE_Components.h if no component is selected (#926)

grasci-arm:
* Do not create RTE_Components.h if no component is selected

* [projmgr] Do not emit reference to `RTE_Components.h` when it does not exist

---------

Co-authored-by: Daniel Brondani <daniel.brondani@arm.com>


jkrech (approved):
LGTM

PR #1528, opened 2024-05-21
[cbuildgen] Implement project-level executes for compatibility with cbuild2cmake

EaselinkBachmann:
This PR implements project-level executes for `cbuildgen` (in addition to it already being implemented for `cbuild2cmake`) since otherwise some projects would depend on users passing `--cbuild2cmake` to compile properly.

The main difference to the implementation in `cbuild2cmake` is the working directory in which the run command is invoked (the project directory instead of the intermediate directory), since the directory would necessarily be different from `cbuild2cmake` (`tmp/project/context` vs `tmp/`) and since the project directory is IMO the correct directory for these commands to be invoked in. For more discussion on that that, see [cmsis-toolbox#132](https://github.com/Open-CMSIS-Pack/cmsis-toolbox/issues/132).

This implementation does not implement access sequences yet, but implements `$input$` and `$output$` for executes nodes.

### DIFF
--- a/tools/buildmgr/cbuild/include/CbuildModel.h
+++ b/tools/buildmgr/cbuild/include/CbuildModel.h
@@ -551,6 +551,26 @@ public:
   */
   const RteTarget* GetTarget() const;
 
+  const std::map<std::string, std::string>& GetExecuteRun() const {
+    return m_executeRun;
+  }
+
+  const std::map<std::string, bool>& GetExecuteAlways() const {
+    return m_executeAlways;
+  }
+
+  const std::map<std::string, std::vector<std::string>>& GetExecuteInputs() const {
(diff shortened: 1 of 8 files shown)


brondani:
`cbuildgen` along with the `CPRJ` format is being deprecated, the tendency is to reject this PR.
https://github.com/Open-CMSIS-Pack/devtools/issues/434

EaselinkBachmann:
> `cbuildgen` along with the `CPRJ` format is being deprecated, the tendency is to reject this PR. #434

That is understandable. I will still maintain the `cbuildgen` executes implementation as long as `cbuildgen` is still the default and not replaced by `cbuild2cmake`, but I understand if you don't want to invest any more development time into a deprecated tool.

PR #1544, opened 2024-05-31
Update IAR.9.32.1.cmake for Secure

jeromecoutant:
Implementation aligned with other cmake files.

@fred-r @tarek-bochkati


jkrech (approved):
Looks good to me. Also checked that the check for the linker is already consistent.

Issue #1266, opened 2024-01-09
Explore usage of Build Information files to provide project information in IDEs


ReinhardKeil:
**Describe The Problem To Be Solved**
There is a risk that the source files of components are not resolved in the same way as the build process is using source files. This could be caused by local pack installations or explicit pack path information.

**Suggest A Solution**
This could be solved by using the csolution build information files that are also used for the CMake process.  In this way we ensure that the IDE has knowledge about the exact source files.

This requires that build information files are also created in case of errors such as components that cannot be resolved or missing generator output files. 

Build Information files may also contain information about generator run commands and potentially other component related information (API, documentation, etc.)

Relates to #1265

jkrech:
@ReinhardKeil  can we close this one in track the implementation (`cbuild setup`) here: https://github.com/Open-CMSIS-Pack/devtools/issues/1265

PR #1564, opened 2024-06-15
[projmgr] Generate RTE directory in project tree

Torbjorn-Svensson:
When running `csolution convert -s /absolute/path/to/foo.csolution.yml`, the "RTE" directory should be placed within that structure and not where the csolution process happens to be executed from.

This is a regression in csolution 2.3.0 compared to 2.2.1 and was introduced by [this change](https://github.com/Open-CMSIS-Pack/devtools/commit/27678e9e3e631fd4e60ee2bf85fd818b8ae328ac#diff-2f6999444987cfc670f1760728ffcdcce7274b4493b1f3b159b1a5719e4409e3L511-L513). In the commit, the RTE directory is now generated without initializing the project, so the path to the project will be the empty string rather than the actual location on disk and as a result, it's created in the working directory rather than in the project tree.

I've added a test case to demonstrate the problem, but I don't know how to fix it.
@jkrech | @edriouk: Can please you take a look?

### DIFF
--- a/libs/rtefsutils/include/RteFsUtils.h
+++ b/libs/rtefsutils/include/RteFsUtils.h
@@ -245,6 +245,13 @@ public:
    * @return absolute path of the current working directory
   */
   static std::string GetCurrentFolder(bool withTrailingSlash = true);
+
+  /**
+   * @brief change current working directory to path
+   * @param path path to use as current working directory
+  */
+  static void SetCurrentFolder(const std::string &path);
+
   /**
    * @brief determine first file owning the given extension in given folder
    * @param folder name of path where first file with given extension is to be found
--- a/libs/rtefsutils/src/RteFsUtils.cpp
+++ b/libs/rtefsutils/src/RteFsUtils.cpp
@@ -594,6 +594,11 @@ string RteFsUtils::GetCurrentFolder(bool withTrailingSlash) {
   return f;
 }
 
+void RteFsUtils::SetCurrentFolder(const string& path) {
+  error_code ec;
+  fs::current_path(path, ec);
+}
+
 bool RteFsUtils::MakeSureFilePath(const string &filePath) {
   fs::path path(filePath);
   path.remove_filename();
--- a/tools/projmgr/test/src/ProjMgrTestEnv.cpp
+++ b/tools/projmgr/test/src/ProjMgrTestEnv.cpp
@@ -50,6 +50,15 @@ StdStreamRedirect::~StdStreamRedirect() {
   std::cerr.rdbuf(m_stdcerrStreamBuf);
 }
 
+TempSwitchCwd::TempSwitchCwd(const string& path) {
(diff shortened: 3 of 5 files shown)


brondani:
Thanks for reporting this. As you correctly says it was introduced in https://github.com/Open-CMSIS-Pack/devtools/pull/1284 addressing https://github.com/Open-CMSIS-Pack/devtools/issues/1266.
Please note it was not a regression: prior to that change the `convert` processing was fully stopped when the first error was found, so nothing was created, while the required change purposely continue the processing also in case of errors. Importantly the return code is different from 0 and the cbuild-idx file brings the `errors: true` node.

As a matter of fact the generation of `RTE_Components.h` was suppressed in https://github.com/Open-CMSIS-Pack/devtools/pull/1527 when no components are selected, so your test case seems not to be valid anymore.
If I am missing something please amend the PR.

Torbjorn-Svensson:
While it's true that nothing was produced earlier, it's still a regression since now, there is a random `RTE` directory created in the directory you launched the `csolution` binary from. Lets say that you are standing in `/path/to/some/read-only/installation/directory/` when you start `csolution`. This would cause other errors to pop-up.
Even worse is that you might stand in a different project and just needs to run `csolution convert` with an absolute path to the other project and now the project in cwd is corrupted.

I would say that there are 2 possible solutions:
1. Only create the RTE directory if project was successfully parsed. (more or less a revert...)
2. Always initialize where the project is located and generate it there.

The current behavior is unpredictable and may introduce other problems for the end-user.

This issue is not about that the identifier that something went wrong has changed, but about that a stray RTE directory can be placed anywhere, not just in the project where it's supposedly to be generated (due to the uninitialized internal state).
From what I can see, this statement needs to always be executed prior to creating the RTE directory: https://github.com/Open-CMSIS-Pack/devtools/blob/main/tools/projmgr/src/ProjMgrWorker.cpp#L556

brondani:
> Only create the RTE directory if project was successfully parsed

This seems to be already the case since https://github.com/Open-CMSIS-Pack/devtools/pull/1527, can you please fix the [unittests building issues reported by CI](https://github.com/Open-CMSIS-Pack/devtools/actions/runs/9530137444) and update the test case?

Torbjorn-Svensson:
> > Only create the RTE directory if project was successfully parsed
> 
> This seems to be already the case since #1527, can you please fix the [unittests building issues reported by CI](https://github.com/Open-CMSIS-Pack/devtools/actions/runs/9530137444) and update the test case?

Updated the test case and it appears to work now.
Not sure how I could have missed this, as I know that I intended to test with `main`, but maybe I forgot to build after I found the commit that caused the problem... Anyway, I think the test case is valuable regardless as it's the only test cases that tries to run `csolution` in a different directory than the .csolution.yml file is located in.

brondani (approved):
LGTM

PR #1579, opened 2024-06-21
[buildmgr] improve relative path support

VGRSTM:
When using specific data structure we may end up with out of range exception (must have fix for 2.5.0 please)


jeromecoutant:
I can't verify the patch... but I confirm many 2.4.0 toolbox users got crash recently when using cbuildgen...

> must have fix for 2.5.0 please

Agree, Thx

brondani (approved):
LGTM

Issue #1590, opened 2024-06-25
Improve error message in case of missing compiler registration


ReinhardKeil:
**The Problem To Be Solved**
When there is no environment variable for [Compiler Registration](https://github.com/ReinhardKeil/cmsis-toolbox/blob/main/docs/installation.md#compiler-registration) the error message does not hint to a solution.

To Duplicate with GCC

```
set GCC_TOOLCHAIN_13_2_1=                // remove compiler registration
cbuild SimpleTZ.csolution.yml --toolchain GCC --update-rte                     
error csolution: toolchain environment variable for 'GCC' was not found
```

**Suggest A Solution**

Change:
`error csolution: toolchain environment variable for 'GCC' was not found` 

To:
`error csolution: compiler registration environment variable missing, format: GCC_TOOLCHAIN_<major>_<minor>_<patch>`

PR #1600, opened 2024-06-27
[projmgr] Further improve warning/error messages (#1590)

Torbjorn-Svensson:
For `csolution list toolchains`, hint that the user needs to define environment variables in case there is no valid toolchain.

For other commands, hint that the environment variable CMSIS_COMPILER_ROOT might need to be defined.


VGRSTM (changes requested):

tools/projmgr/src/ProjMgr.cpp
```diff
@@ -1005,25 +1005,30 @@ bool ProjMgr::RunListToolchains(void) {
...
-      string env = toolchain.version;
-      replace(env.begin(), env.end(), '.', '_');
-      if (!toolchain.root.empty()) {
-        toolchainEntry += "  Environment: " + toolchain.name + "_TOOLCHAIN_" + env + "\n";
-        toolchainEntry += "  Toolchain: " + toolchain.root + "\n";
-      }
-      if (!toolchain.config.empty()) {
-        toolchainEntry += "  Configuration: " + toolchain.config + "\n";
+  if (toolchains.empty()) {
+    cerr << "No valid compiler registered. Consider to define the CMSIS_COMPILER_ROOT environment variable and compiler registration environment variables, format: <GCC|AC6|IAR>_TOOLCHAIN_<major>_<minor>_<patch>\n";
```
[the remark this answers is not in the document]
Torbjorn-Svensson: Didn't think of that, fixed.

tools/projmgr/src/ProjMgr.cpp
```diff
@@ -1005,25 +1005,30 @@ bool ProjMgr::RunListToolchains(void) {
...
-      string env = toolchain.version;
-      replace(env.begin(), env.end(), '.', '_');
-      if (!toolchain.root.empty()) {
-        toolchainEntry += "  Environment: " + toolchain.name + "_TOOLCHAIN_" + env + "\n";
-        toolchainEntry += "  Toolchain: " + toolchain.root + "\n";
-      }
-      if (!toolchain.config.empty()) {
-        toolchainEntry += "  Configuration: " + toolchain.config + "\n";
+  if (toolchains.empty()) {
+    cerr << "No valid compiler registered. Consider to define the CMSIS_COMPILER_ROOT environment variable and compiler registration environment variables, format: <GCC|CLANG|AC6|IAR>_TOOLCHAIN_<major>_<minor>_<patch>\n";
```
brondani, line 1009: The `CMSIS_COMPILER_ROOT` is optional as documented in the [installation guide](https://github.com/Open-CMSIS-Pack/cmsis-toolbox/blob/main/docs/installation.md#environment-variables).
It would be useful to understand in which scenario this is mandatory for you, could you please give us more information on this matter?

Since this regards the generic call of `list toolchains`, without specifying csolution.yml, a more appropriate location to introduce such error message could be here:
https://github.com/Open-CMSIS-Pack/devtools/blob/tools/projmgr/2.4.0/tools/projmgr/src/ProjMgrWorker.cpp#L4050

Other observations:
- Please use `ProjMgrLogger::Error` instead of directly writing to `cerr`.
- For consistency please write error messages in a single sentence without final dot.

Torbjorn-Svensson: I know that the `CMSIS_COMPILER_ROOT` is optional, but in some cases, it's an alone csolution binary and in that case, the user needs some hints to what is needed to get it running. Maybe the distribution was missing some files, maybe there is no compatible version of the files in the `etc` directory, regardless, there should be some help to get it working.

I was considering putting the block in the ProjMgrWorker class, but as there are other output in the ProjMgr class in this call hierarchy, I though it made more sense to have it there. Do you still want me to move it?

Also, the reason to call `cerr` was that `cout` is used in the other branch of the function in ProjMgr class.

Let me know how you want to proceed.
Once I have the reply to the above statements, I will try to rephrase the sentences a bit.

brondani: > I know that the CMSIS_COMPILER_ROOT is optional, but in some cases, it's an alone csolution binary and in that case, the user needs some hints to what is needed to get it running. Maybe the distribution was missing some files, maybe there is no compatible version of the files in the etc directory, regardless, there should be some help to get it working.

Then we would need more separated warnings/errors in different places TBD:
- no toolchain environment variables
- `<cmsis-toolbox>/etc` directory does not exist and `CMSIS_COMPILER_ROOT` was not set or is invalid
- no compatible toolchain config files

It's fine for me if you want to proceed with this PR without addressing those now but in my opinion the most useful would be to inform about a missing `etc` directory that is a symptom of a malformed installation. 
(@ReinhardKeil @jkrech please advise)

> I was considering putting the block in the ProjMgrWorker class, but as there are other output in the ProjMgr class in this call hierarchy, I though it made more sense to have it there. Do you still want me to move it?

Yes please.

> Also, the reason to call cerr was that cout is used in the other branch of the function in ProjMgr class.

`cout` is used for outputing results of `list` commands.
All messages (info, warning, error, debug) should use `ProjMgrLogger`.

Thanks!

jkrech: `csolution` is not expected to work without  `<cmsis-toolbox>/etc` being present. Where `<cmsis-toolbox>/bin` specifies the path of the running csolution binary itself.
In the etc/ directory `csolution` expects schema files and the `global.generator.yml`.

=> error message if `../etc` does not exist
=> error message in schema check when a schema file is missing from `<cmsis-toolbox>/etc`
=> error when not finding a matching toolchain configuration file `<toolchain>.<minVersion>.cmake`. 
IF $CMSIS_COMPILER_ROOT is set, then the message reads: not found in `$CMSIS_COMPILER_ROOT: <cmsis_compiler_root_value>`
ELSE
the message reads: not found in `<cmsis-toolbox>/etc: <cmsis-toolbox-value>/etc`

This looks like three different error messages in three different points in the execution.

Torbjorn-Svensson: Sorry, but I will not have time to do this in the near future.
Is this something that Arm can complete or should it be postponed?

jkrech: @Torbjorn-Svensson is that something you would have time to work on in the next couple of weeks? 
If so, please consider the message suggested in the last comment of #1728.

Torbjorn-Svensson:
@jkrech I've rebased the PR and updated the text with the addition of CMSIS_COMPILER_ROOT that I think is missing in the text shown in 2.5.0.

soumeh01 (commented):
@Torbjorn-Svensson I have updated the error text as decided in https://github.com/Open-CMSIS-Pack/devtools/issues/1728

Changes are under commit:

- [Updated error message](https://github.com/Open-CMSIS-Pack/devtools/pull/1600/commits/c1c54507f6925f7bb902c4ca3b578c145fe6d3bc)
- [Updated tests](https://github.com/Open-CMSIS-Pack/devtools/pull/1600/commits/64cfc5546720bddea09496e2fd051456cb08b299)
- [Updated test](https://github.com/Open-CMSIS-Pack/devtools/pull/1600/commits/e12fe13a11d7829f46799c3e63820bd3278a8fbf)

Torbjorn-Svensson:
Well, the thing I want to stress once more is that CMSIS_COMPILER_ROOT may not contain what is expected and then the message that is printed won't contain the clue that it's not *_TOOLCHAIN_major_minor_patch that is missing, but instead the .cmake file for the toolchain.

Example:
```
$ mkdir /tmp/foobar
$ AC6_TOOLCHAIN_7_2_3=/bin CMSIS_COMPILER_ROOT=/tmp/foobar csolution list toolchains
error csolution: no compiler registered. Add path to compiler ./bin directory with environment variable <name>_TOOLCHAIN_<major>_<minor>_<patch>. <name> is one of AC6, GCC, IAR, CLANG
```

In the above example, the CMSIS_COMPILER_ROOT does not contain any AC6 cmake file, but there is a toolchain "registered" using the environment variable. With this scenario, the error text is simply not helpful.

jkrech:
@Torbjorn-Svensson 
Since the registration of toolchains using environment variables I think the error message above is sufficient.
Once you have registered the toolchain required by the solution we can then check whether a corresponding compiler.cmake file is present matching the compiler version requirements. If that is not found and no CMSIS_COMPILER_ROOT is set, then the installation is incomplete and the message should point to the installation's ./etc directory. If CMSIS_COMPILER_ROOT is set but no cmake file is found it the message should print the CMSIS_COMPILER_ROOT variable name and value.

Torbjorn-Svensson:
So, consider the following example then:

CMSIS_COMPILER_ROOT is populated with a file named AC6.6.18.0.cmake
There are no \*\_TOOLCHAIN\_\* environment variables.
```
$ AC6_1_2_3=/bin csolution list toolchains
```
In this case, would you expect there to be the same error message or something else?
Note, the available toolchains are a combination of the content of CMSIS_COMPILER_ROOT and the \*\_TOOLCHAIN\_\* environment variables.

Again, as a user, there is nothing helping me to understand what is missing with the error message like `error csolution: no compiler registered. Add path to compiler ./bin directory with environment variable <name>_TOOLCHAIN_<major>_<minor>_<patch>. <name> is one of AC6, GCC, IAR, CLANG`.
It would be helpful to get the clue that it's not only the \*\_TOOLCHAIN\_\* environment variables that are needed here.

The important part in the example above is that the toolchain is specified, but it's considered "invalid" by the csolution utility for a reason that might not be obvious to the user.

ReinhardKeil:
`Add path to compiler ./bin directory with environment variable` is IMHO pretty clear.

I accept that we cannot cover every aspect in a 1-line error message.  We have therefore also a user's guide:
https://github.com/Open-CMSIS-Pack/cmsis-toolbox/blob/main/docs/installation.md

Torbjorn-Svensson:
Okay, so is this PR okay after the updates?
Feel free to merge.

jkrech (approved):
LGTM

brondani (approved):

PR #1611, opened 2024-07-01
Error on use of multiple variants of the same component for the same context (#960)

grasci-arm:
* Error on use of multiple variants of the same component for the same context

* check for multiple component variants


jkrech (approved):
LGTM

Issue #1641, opened 2024-07-15
Improve documentation on `cdefault.yml`


ReinhardKeil:
**The Problem To Be Solved**

`cdefault.yml` is a good starting point, but there are a few issues that should be improved for **GCC**.

1. Link option `--specs=rdimon.specs` 
     Enables semihosting, but may actually create problems as it runs on a breakpoint. Should this be in the default?

2. Linker warning `warning: ...elf has a LOAD segment with RWX permissions` when using the default linker script management.
     What is causing this? Can it be solved or does this need to be in the documentation?

jkrech:
The rational behind enabling semihosting for GCC was driven by running solutions on AVH-FVP models using semihosting for test output.

ReinhardKeil:
Decision on 1. is to not enable semihosting (leave the existing toolchain default).

Issue #1576, opened 2024-06-20
[cbuild setup] list selectable compiler for a solution without specified `compiler:` node based on  cdefault and registered toolchains


jkrech:
**Describe The Problem To Be Solved**
Provide a list of selectable compiler (versions)  for a toolchain agnostic CMSIS solution in `*.cbuild-idx.yml` based on "compilers listed in cdefault" and the compiler versions registered via environment variables.

** Suggest A Solution**
 
Extend cbuild setup command to deliver available toolchains for a project in the cbuild-idx.yml file when no toolchain is selected. The node is [select-toolchains:]([https://github.com/Open-CMSIS-Pack/cmsis-toolbox/blob/main/docs/YML-CBuild-Format.md#select-toolchains]](https://github.com/Open-CMSIS-Pack/cmsis-toolbox/blob/main/docs/YML-CBuild-Format.md#select-toolchains%5D).

Refer to [cbuild setup command](https://github.com/Open-CMSIS-Pack/cmsis-toolbox/blob/main/docs/build-operation.md#cbuild-setup-command) for details.

Possible Implications

    no longer use (or allow) compiler: selection in cdefault.yml for toolchain agnostic projects.

Related proposal:
https://github.com/Open-CMSIS-Pack/devtools/issues/1507

ReinhardKeil:
Process is slightly changed, see: https://github.com/ReinhardKeil/cmsis-toolbox/blob/main/docs/YML-Input-Format.md#select-compiler

jkrech:
@ReinhardKeil can you please clarify that the `select-compiler` can only be used in `cdefault.yml` and `csolution.yml` files or is my understanding wrong?

brondani:
Implemented in `csolution convert` in https://github.com/Open-CMSIS-Pack/devtools/pull/1581.

jkrech:
using csolution nightly 2024-06-24:
*.csolution.yml:
```yml
...
  select-compiler:
    - compiler: AC6
    - compiler: GCC
    - compiler: CLANG
    - compiler: IAR
 ...
```

Registered Toolchains are
```
>cbuild list toolchains
AC6@6.22.0
CLANG@18.1.3
GCC@13.2.1
```

running command:
```
cbuild Hello.csolution.yml 
error csolution: processing context 'Hello.Debug+AVH' failed
error csolution: processing context 'Hello.Release+AVH' failed
[...]

PR #1643, opened 2024-07-16
Update cdefault.yml

jkrech:
#1641 do not enable semihosting by default, as this will not work for hardware. #1576 Do not set `compiler:` node in cdefault to trigger compiler detection for solutions without selected compiler when cdefault from etc is used.

### DIFF
--- a/tools/projmgr/templates/cdefault.yml
+++ b/tools/projmgr/templates/cdefault.yml
@@ -1,7 +1,5 @@
 default:
 
-  compiler: GCC
-
   misc:
     - for-compiler: AC6
       C-CPP:
@@ -30,7 +28,6 @@ default:
         - -std=gnu11
       Link:
         - --specs=nano.specs
-        - --specs=rdimon.specs
         - -Wl,-Map=$elf()$.map
         - -Wl,--gc-sections
 


tools/projmgr/templates/cdefault.yml
```diff
@@ -1,7 +1,5 @@
 default:
 
-  compiler: GCC
```
JonatanAntoni, line 3: This may break existing projects not setting a default compiler elsewhere.

tools/projmgr/templates/cdefault.yml
```diff
@@ -30,7 +28,6 @@ default:
         - -std=gnu11
       Link:
         - --specs=nano.specs
-        - --specs=rdimon.specs
```
JonatanAntoni, line 33: This may break existing projects using semihosted stdout for printing analysis or test results.

JonatanAntoni (approved):
Changes are not backward compatible.

brondani (dismissed):

ReinhardKeil:
> Changes are not backward compatible.

That's understood and needs to be documented.

tools/projmgr/templates/cdefault.yml
jkrech: Understood, this will cause some friction in project using the cdefault from etc.

tools/projmgr/templates/cdefault.yml
jkrech: Agreed, I think there is a clear error message indicating what is missing. 
`error csolution: compiler undefined, use '--toolchain' option or add 'compiler: <value>' to yml input`

brondani (approved):

tools/projmgr/templates/cdefault.yml
```diff
@@ -30,9 +28,9 @@ default:
...
         - -Wl,-Map=$elf()$.map
         - -Wl,--gc-sections
+        - -Wl,--no-warn-rwx-segments
```
Torbjorn-Svensson, line 33: I think this change is dangerous as it would silence the warning that the application is open for a wide range of possible attacks.
Segments in an .elf file that requires both executable and write permission at the same time is vulnerable to a overflow attack and would allow to overwrite the application code in these segments to inject custom code into the application space and potentially leak sensitive information or take over the application (ransom attack for example).

ReinhardKeil:
The trigger point is this issue: https://github.com/Open-CMSIS-Pack/devtools/issues/1641 #2. 
Do you have a solution to the underlying linker problem that is explained in more detail here:
https://metebalci.com/blog/solution-to-load-segment-with-rwx-permissions-warning-with-cmsis/

Note that the warning itself does not prevent from loading and executing code into RAM.  Therefore it is not a solution to ransom attacks.  That could be achieve i.e. by using the MPU.

tools/projmgr/templates/cdefault.yml
```diff
@@ -30,9 +28,9 @@ default:
         - -std=gnu11
       Link:
         - --specs=nano.specs
-        - --specs=rdimon.specs
```
Torbjorn-Svensson, line 33: If you drop `--specs=rdimon.specs`, you likely need to add `--specs=nosys.specs` as you need some implementation of the syscalls functions.
The nosys.specs contains the default implementation (returning error) wile rdimon.specs contains a full semihosting implementation of the same syscalls.
If you have neither of them, you either need to implement the used syscalls yourself or you will get undefined symbol linker error.

Torbjorn-Svensson:
> @Torbjorn-Svensson The trigger point is this issue: #1641 item 2. Do you have a solution to the underlying linker problem that is explained in more detail here: https://metebalci.com/blog/solution-to-load-segment-with-rwx-permissions-warning-with-cmsis/
> 

I don't know the reason why the `.copy.table` symbol is flagged as read-write instead of just read.
In our linker scripts, we have marked sections containing pure data as `(READONLY)`, but it's only available binutils 2.38 and later versions (was first available in 11.3.rel1 of the Arm released GCC builds for arm-none-eabi).

> Note that the warning itself does not prevent from loading and executing code into RAM. Therefore it is not a solution to ransom attacks. That could be achieve i.e. by using the MPU.

Absolutely, keeping the warning does not fix the issue, but at least the user is aware that the issue exist.

ReinhardKeil:
@Torbjorn-Svensson thanks for all your feedback.

Let me explain the situation with the warning: today users get unfortunately an incorrect warning.  We did ask our internal GCC team to deliver a solution - unfortunately this is not available yet.  Hence the proposal is to disable the incorrect warning for the time being (until the fundamental problem has been solved).

Note: If you have a solution for the fundamental problem, that would be great.

PR #1648, opened 2024-07-16
Update cdefault.yml

jkrech:
remove semihosting for CLANG and IAR as well.

@JonatanAntoni could you please validate that this works correctly for CLANG and IAR.

### DIFF
--- a/tools/projmgr/templates/cdefault.yml
+++ b/tools/projmgr/templates/cdefault.yml
@@ -40,8 +40,7 @@ default:
       C:
         - -std=gnu11
       Link:
-        - -lcrt0-semihost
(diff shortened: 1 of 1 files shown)


JonatanAntoni (approved):
LGTM

JonatanAntoni:
Sorry, I tried to run GCC and Clang for CoreValidation (I cannot do it for IAR):

GCC fails with just `--specs=nano.specs` because it is missing lots of functions. Instead, `--specs=nosys.specs` works.

Clang fails with just `-lcrt0` because `stdout` and `_exit` are missing.

JonatanAntoni (approved):

Issue #1670, opened 2024-07-31
Improve generated linker `regions*.h` file


ReinhardKeil:
**The Problem To Be Solved**

The linker `regions*.h` uses `< >` sequences that may conflict with config wizard annotations.  Also the output could be slightly improved as suggested below.

**(Optional): Suggest A Solution**

Change output to (note that BSP is actually not used in this example - just a mockup).
```
//-------- <<< Use Configuration Wizard in Context Menu >>> -------

// <n>Generated by csolution 2.5.0 using information from packs
// <i>Device Family Pack (DFP):   Keil::LPC1700_DFP@2.7.2
// <i>Board Support Pack (BSP):   Keil::MCB1700_BSP@2.3.0

// <h>ROM Configuration
// =======================
// <h> ROM0 (is IROM1 from DFP)
//   <o> Base address <0x0-0xFFFFFFFF:8>
//   <i> Defines base address of memory region.
//   <i> Default: 0x00000000
#define __ROM0_BASE 0x00000000
//   <o> Region size [bytes] <0x0-0xFFFFFFFF:8>
//   <i> Defines size of memory region.
//   <i> Default: 0x80000
#define __ROM0_SIZE 0x80000
//   <q>Default region
//   <i> Enables memory region globally for the application.
#define __ROM0_DEFAULT 1
//   <q>Startup
//   <i> Selects region to be used for startup code.
#define __ROM0_STARTUP 1
// </h>

// </h>

// <h>RAM Configuration
// =======================
// <h> RAM0 (is IRAM1 from DFP)
//   <o> Base address <0x0-0xFFFFFFFF:8>
//   <i> Defines base address of memory region.
//   <i> Default: 0x10000000
#define __RAM0_BASE 0x10000000
//   <o> Region size [bytes] <0x0-0xFFFFFFFF:8>
//   <i> Defines size of memory region.
//   <i> Default: 0x8000
#define __RAM0_SIZE 0x8000
//   <q>Default region
//   <i> Enables memory region globally for the application.
#define __RAM0_DEFAULT 1
//   <q>No zero initialize
//   <i> Excludes region from zero initialization.
#define __RAM0_NOINIT 0
// </h>

// <h> RAM1 (is IRAM2 from DFP)
//   <o> Base address <0x0-0xFFFFFFFF:8>
//   <i> Defines base address of memory region.
//   <i> Default: 0x2007C000
#define __RAM1_BASE 0x2007C000
[...]

PR #1672, opened 2024-07-31
avoid special characters in regions.h confusing config wizard parsing

jkrech:
Adressing: https://github.com/Open-CMSIS-Pack/devtools/issues/1670


ReinhardKeil:
Question: are we today not adding memory information from BSPs?

jkrech:
Revised the output to differentiate between DFP and BSP.

edriouk (approved):

PR #1713, opened 2024-08-22
Avoid creation of files when dry-running generator

ErikMallbergSTM:
Both generator output directory and RTE were created (in the latter case also with contents) when running a generator with the `--dry-run` flag. The idea of dry run was to only return the contents of the GPDSC without unnecessary work or file system access.

When the RTE directory is no longer updated when dry-running a generator, creation of the cbuild and cbuild-gen files need to ignore that files may be missing when adding the `constructed-files` node.


brondani (approved):
LGTM

PR #1755, opened 2024-09-11
Delete templates {{ProjectName}}.cproject.yml and {{SolutionName}}.csolution.yml

jkrech:
superseded by template projects


brondani (approved):
Please note it also needs to be removed from cmsis-toolbox tests:
https://github.com/Open-CMSIS-Pack/cmsis-toolbox/blob/main/test/tests.py#L19-L20

PR #1807, opened 2024-10-14
buildmgr: Patch CLANG.cmake to be compatible with LLVM 19

JonatanAntoni:
In LLVM 19 the order of the target triple got swapped. I.e., none-unknown-eabi is not accepted anymore instead unknown-none-eabi is required. By dropping the unknown entirely the reduced target triple is accepted by all major versions 17, 18, and 19 by defaulting to vendor unknown.


brondani (approved):
LGTM

Issue #1057, opened 2023-07-11
[csolution] cdefault: GCC inline assembly shall accept unified syntax by default 


JonatanAntoni:
**Describe The Problem To Be Solved**

For inline assembly we typically use unified syntax to be as compiler-agnostic as possible. GCC still uses disjoint (i.e., pre-unified) syntax by default for history reasons. This could make compilation units using unified syntax fail to compile.

**Suggest A Solution**

GCC has the command line flag `-masm-syntax-unified` to switch to unified syntax. This flag shall be used by default, e.g., by adding it into `cdefault.yml`.

JonatanAntoni:
Fixed in #1045.

PR #1835, opened 2024-10-31
Workflows security updates (#1057)

spcaipers-arm:


jkrech (approved):
LGTM
What is the expected score improvement?
How much more effort does this mean for maintaining the versions?

PR #1848, opened 2024-11-13
Updated version string

spcaipers-arm:


jkrech:
@thorstendb-ARM  do you see any problems with these changes to the svdconv output messages?

PR #1972, opened 2025-02-18
Debug Config file PLM

spcaipers-arm:


libs/rtemodel/include/RteItem.h
```diff
@@ -787,6 +798,18 @@ class RteItem : public XmlTreeItem<RteItem>
   */
   virtual RteItem* GetLicenseSet() const;
 
+  /**
+   * @brief check if this item represents or is associated with a that is a config one
+   * @return true if "attr" attribute value is "config" in default implementation
+  */
+  virtual bool IsConfig() const;
```
brondani, line 805: @edriouk Please also note the macos build is throwing this warning:
```
'IsConfig' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
```

Issue #1733, opened 2024-09-02
[packchk] check if the device's memory attribute `startup`, `uninit`, etc. are present for all devices


ReinhardKeil:
**The Problem To Be Solved**

Automatic linker script generation relies on `startup` and `uninit` memory attributes.

Decide if we need to enhance packchk to report missing/duplicate attributes for memory regions such as:

- when no rx region has startup
- when no rw region has uninit
- when 2 or more rx regions have startup
- when 2 or more rw regions have uninit

thorstendb-ARM:
Discussed with @ReinhardKeil , implementing the "startup" checks only.
Check: Gather all "startup=1", test "x" attribute. At the end, check if exactly one was found.

thorstendb-ARM:
Changed: Implicily assume attribute eXecution for IROMx / IRAMx.

PR #1980, opened 2025-02-20
[packchk] check if the device's memory attribute startup, uninit, etc. are present for all devices #1733 (#1153)

grasci-arm:


edriouk (approved):

RTE/_Test/RTE_Components.h
brondani, line 1: @thorstendb-ARM @edriouk 
Was this file committed on purpose? If not, please clean it up.

PR #2075, opened 2025-05-09
[cbuild-run] Fix effective `datapatch` handling

grasci-arm:


jkrech (approved):
LGTM - added test case matches

PR #2094, opened 2025-05-28
Do not validate for http:// URLs in elements and attributes that are …

### LINKED ISSUE #2093: [packchk] Namespaces does not need to be https URLs
**Describe the bug**
In the latest packchk 2.9.0, we are now triggering a warning M300

```
*** WARNING M300: /home/jenkins/agent/workspace/dfp/DFP-PIC32CM-MC/pack-packchk/build/Microchip.PIC32CM-MC_DFP/1.5.204/Microchip.PIC32CM-MC_DFP.pdsc (Line 776) 
  at must use 'https://'
```

This is caused by the .pdsc fragment 

```xml
<at:extension xmlns:at="http://www.atmel.com/schemas/pack-device-atmel-extension"
                             schemaVersion="1.0"
                             xsi:schemaLocation="http://www.atmel.com/schemas/pack-device-atmel-extension PACK.DEVICE.ATMEL.EXTENSION.xsd"><!--https://schemas.microchip.com/com/microchip/packs/extensions/at/1.0/PACK.DEVICE.ATMEL.EXTENSION.xsd-->
```

The warning also triggers on the `schemaLocation`, which granted is correct, a schema should only load over HTTPS. However the namespace URL itself is meaningless, as it is just a identifier that is a URL by convention. It is also not changeable without breaking the namespace consumers altogether.

**To Reproduce**

Run `packchk -i ARM.CMSIS.pdsc Microchip.PIC32CM-MC_DFP.pdsc`

[Microchip.PIC32CM-MC_DFP-1.5.204.pdsc.zip](https://github.com/user-attachments/files/20490569/Microchip.PIC32CM-MC_DFP-1.5.204.pdsc.zip)

**Expected behavior**
Warning should be emitted for `schemaLocation`, but not for namespace URL.

**Environment (please complete the following information):**

- packchk
- Version 2.9.0
- OS: linux

xoriath:
…definitely part of a different namespace

Since the XML parser does not seem to have much support at all for namespaces, this fix is very ... raw in that it only looks for a namespace delimiter `:` and if we find that in an element, mark it as ok. If we find it in an attribute, do not check it.

This fixes #2093

### DIFF
--- a/tools/packchk/src/CheckFiles.cpp
+++ b/tools/packchk/src/CheckFiles.cpp
@@ -40,6 +40,10 @@ CheckFilesVisitor::~CheckFilesVisitor()
 */
 VISIT_RESULT CheckFilesVisitor::Visit(RteItem* item)
 {
+  const string& tag = item->GetTag();
+  if(tag == "environment") {
+    return VISIT_RESULT::SKIP_CHILDREN;
+  }
   m_checkFiles.CheckFile(item);
   m_checkFiles.CheckUrls(item);
   m_checkFiles.CheckDeprecated(item);
--- a/tools/packchk/test/data/TestUrlHttps/TestVendor.TestUrlHttps.pdsc
+++ b/tools/packchk/test/data/TestUrlHttps/TestVendor.TestUrlHttps.pdsc
@@ -47,6 +47,13 @@
       <subFamily DsubFamily="MyDevice">
         <device Dname="MyDeviceM0">
           <processor Dfpu="0" Dmpu="0" Dendian="Little-endian" Dclock="24000000"/>
+          <environment name="test_env">
+            <e:extension xmlns:e="http://example.com"
+                            schemaVersion="1.0"
+                            xsi:schemaLocation="http://example.com example.xsd">
+               <e:element value="test"/>
+            </e:extension>
+         </environment>
         </device>
       </subFamily>
     </family>
--- a/tools/packchk/test/integtests/src/PackChkIntegTests.cpp
+++ b/tools/packchk/test/integtests/src/PackChkIntegTests.cpp
@@ -806,7 +806,7 @@ TEST_F(PackChkIntegTests, CheckUrlForHttp) {
   }
 
   if (M300_foundCnt != 8) {
-    FAIL() << "error: missing message M300";
+    FAIL() << "error: missing number of M300, found " << M300_foundCnt;
   }
 }
 


thorstendb-ARM (dismissed):
Changes seem to fix the issue for Atmel/Microchip, but not sure about if this is a too generic approach.

I'd suggest the following:
```
/**
 * @brief visitor callback for files found in PDSC description
 * @param item RteItem file item
 * @return VISIT_RESULT
*/
VISIT_RESULT CheckFilesVisitor::Visit(RteItem* item)
{
  const string& tag = item->GetTag();
  if(tag == "environment") {
    return VISIT_RESULT::SKIP_CHILDREN;
  }

  m_checkFiles.CheckFile(item);
  m_checkFiles.CheckUrls(item);
  m_checkFiles.CheckDeprecated(item);
  m_checkFiles.CheckDescription(item);
  m_checkFiles.GatherIncPathVsAttrConfig(item);
  m_checkFiles.CheckCSolutionEntries(item);


  return VISIT_RESULT::CONTINUE_VISIT;
}
```

Please also change the test file to be more generic (no references to an existing vendor).

tools/packchk/src/CheckFiles.cpp
```diff
@@ -255,6 +255,9 @@ bool CheckFiles::CheckUrls(RteItem* item)
     
   const auto& attributes = item->GetAttributes();
   for(const auto& [attr, text] : attributes) {
+    if(attr.find(":", 0) != string::npos) {
+      continue;
+    }
     if(text.find("http://", 0) != string::npos) {
```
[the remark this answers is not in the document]
edriouk: Sorry, misinterpreted

tools/packchk/src/CheckFiles.cpp
```diff
@@ -255,6 +255,9 @@ bool CheckFiles::CheckUrls(RteItem* item)
     
   const auto& attributes = item->GetAttributes();
   for(const auto& [attr, text] : attributes) {
+    if(attr.find(":", 0) != string::npos) {
```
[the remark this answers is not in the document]
edriouk: Sorry, misinterpreted

thorstendb-ARM (commented):
.

xoriath:
> Changes seem to fix the issue for Atmel/Microchip, but not sure about if this is a too generic approach.
> 
> I'd suggest the following:
> 
> ```
> /**
>  * @brief visitor callback for files found in PDSC description
>  * @param item RteItem file item
>  * @return VISIT_RESULT
> */
> VISIT_RESULT CheckFilesVisitor::Visit(RteItem* item)
> {
>   const string& tag = item->GetTag();
>   if(tag == "environment") {
>     return VISIT_RESULT::SKIP_CHILDREN;
>   }
> 
>   m_checkFiles.CheckFile(item);
>   m_checkFiles.CheckUrls(item);
>   m_checkFiles.CheckDeprecated(item);
>   m_checkFiles.CheckDescription(item);
>   m_checkFiles.GatherIncPathVsAttrConfig(item);
>   m_checkFiles.CheckCSolutionEntries(item);
> 
> 
>   return VISIT_RESULT::CONTINUE_VISIT;
> }
> ```
> 
> Please also change the test file to be more generic (no references to an existing vendor).

@thorstendb-ARM So this is not generic enough... A namespace can be declared at any level of the DOM, and thus there might be other url-like things... 

Now, maybe the more correct thing is to just not parse anything namespace-like into the RTE model?

tools/packchk/test/data/TestUrlHttps/TestVendor.TestUrlHttps.pdsc
```diff
@@ -47,6 +47,13 @@
       <subFamily DsubFamily="MyDevice">
         <device Dname="MyDeviceM0">
           <processor Dfpu="0" Dmpu="0" Dendian="Little-endian" Dclock="24000000"/>
+          <environment name="test_env">
+            <e:extension xmlns:e="http://example.com"
+                            schemaVersion="1.0"
+                            xsi:schemaLocation="http://example.com example.xsd">
+               <e:element value="test"/>
+            </mchp:extension>
```
edriouk, line 55:
```suggestion
            </e:extension>
```

tools/packchk/src/CheckFiles.cpp
```diff
@@ -241,7 +245,7 @@ bool CheckFiles::CheckDescription(RteItem* item)
 */
 bool CheckFiles::CheckUrls(RteItem* item)
 {
-  if(!item || item->GetTag() == "package") {
+  if(!item || item->GetTag() == "package" || item->GetTag().find(":", 0) != string::npos) {
```
edriouk, line 248: I believe this is a redundant change, code above at lines 44-46 effectively disables all checks for environment-specific elements

edriouk (changes requested):

tools/packchk/test/data/TestUrlHttps/TestVendor.TestUrlHttps.pdsc
thorstendb-ARM: We should not implement URL checks filtering based on any namespace characters. As @edriouk mentioned, my code suggestion should set up the correct filter to not run any checks on vendor specific added content.

thorstendb-ARM:
> @thorstendb-ARM So this is not generic enough... A namespace can be declared at any level of the DOM, and thus there might be other url-like things...

The CMSIS Pack Description format does (currently) not use any namespaces. The extension point `<environment>` allows pack vendors to hook in to add data for other tools. The PackChk does verify some entries for environments "UV" and "DS5", but does not touch others. Thus, vendors are free to put whatever they like here. Most of the checks of PackChk should exclude anything below environment as it is not part of the CMSIS-Pack specification. This is what my code proposal does for this module.

> Now, maybe the more correct thing is to just not parse anything namespace-like into the RTE model?
The RTE Model is written as a generic library, it reads everything that is in the pdsc file and presents it to a consumer.

xoriath:
The thing that still bugs me is, the checking of xmlns: attributes which are allowed to be on any element which will declare a namespace is still being validated and forced to be https, which is really not what the original check intended to verify... But just ignoring most environments works for me specifically so good enough for me 😉

Issue #2128, opened 2025-07-01
GCC build fails when path contains blank character


ReinhardKeil:
**Describe the bug**

This project does not build when the directory path contains a blank character.
[Jun 27.zip](https://github.com/user-attachments/files/20998635/Jun.27.zip)

**To Reproduce**

```bash
C:\Test\Jun 27\Blinky>cbuild Blinky_Solution.csolution.yml --rebuild --context-set   
+--------------------------------------------------------
(1/1) Cleaning context: "Blinky.Debug+PIC32CM5164JH00100"
+--------------------------------------------------------
(1/1) Building context: "Blinky.Debug+PIC32CM5164JH00100"
Using GCC V14.2.1 compiler, from: 'c:/Users/reikei01/.vcpkg/artifacts/2139c4c6/compilers.arm.arm.none.eabi.gcc/14.2.1/bin'
Building CMake target 'Blinky.Debug+PIC32CM5164JH00100'
[1/4] Building C object CMakeFiles/Group_Source_Files.dir/C_/Test/Jun_27/Blinky/Blinky/main.c.obj
[2/4] Building C object CMakeFiles/Microchip_Device_Startup_1_4_0.dir/C_/Test/Jun_27/Blinky/Blinky/RTE/Device/PIC32CM5164JH00100/system_pic32cm5164jh00100.c.obj
[3/4] Building C object CMakeFiles/Microchip_Device_Startup_1_4_0.dir/C_/Test/Jun_27/Blinky/Blinky/RTE/Device/PIC32CM5164JH00100/startup_pic32cm5164jh00100.c.obj
[4/4] Linking C executable "C:\Test\Jun 27\Blinky\out\Blinky\PIC32CM5164JH00100\Debug\Blinky.elf"
FAILED: C:/Test/Jun 27/Blinky/out/Blinky/PIC32CM5164JH00100/Debug/Blinky.elf
[...]

PR #2129, opened 2025-07-01
fix map file with space in path for GCC

jkrech:
fix: #2128


jkrech:
Very valid point. I have updated it for the other compilers as well.

PR #2149, opened 2025-07-18
Added Infineon Device Configurator

KeilChris:


tools/projmgr/templates/global.generator.yml
```diff
@@ -9,3 +9,8 @@ generator:
     download-url: https://www.nxp.com/configtools
     run: ../bin/launch-MCUXpressoConfigTools
     path: $SolutionDir()$/MCUXpressoConfig/$TargetType$
+  - id: InfineonDeviceConfigurator
+    description: Global Registered Generator for Infineon Device Configuration tool
+    download-url: https://softwaretools.infineon.com/tools/com.ifx.tb.tool.deviceconfigurator
+    run: ../bin/launch-Infineon_Dev_Config
+    path: $SolutionDir()$/$Project$/RTE/Device_Configurator
```
jkrech, line 16: Regarding `path:`
Please replace `$SolutionDir()$/$Project$` with `$ProjectDir()$ as we must not make the assumption that a project is always in the sub directory named after the project.

I am also worried  that we do not use a different sub directory for each `$TargetType$`, as this is a device configurator which will generate different files for different devices.
If two target-types have different boards/devices selected the files should be managed separately.

PR #2310, opened 2025-12-15
Add support for Microchip XC32 toolchain in the CMSIS toolbox

ashish-mahanth:
This PR aims to add support for the [XC32 Toolchain](https://www.microchip.com/en-us/tools-resources/develop/mplab-xc-compilers/xc32) and provide a solution for the issue found [here](https://github.com/Open-CMSIS-Pack/cmsis-toolbox/issues/467).

### DIFF
--- a/tools/buildmgr/cbuildgen/config/XC.5.0.0.cmake
+++ b/tools/buildmgr/cbuildgen/config/XC.5.0.0.cmake
@@ -53,7 +53,7 @@ function(cbuild_set_options_flags lang optimize debug warnings language flags)
 endfunction()
 
 set(MDFP "")
-file(GLOB_RECURSE ALL_DIRS LIST_DIRECTORIES true "${DPACK_DIR}")
+file(GLOB_RECURSE ALL_DIRS LIST_DIRECTORIES true "${DPACK_DIR}/*")
 foreach(DIR_PATH ${ALL_DIRS})
     if(IS_DIRECTORY "${DIR_PATH}")
 	    get_filename_component(CURRENT_DIR_NAME "${DIR_PATH}" NAME)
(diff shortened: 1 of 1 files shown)


brondani:
@ashish-mahanth
> Is there a variable which holds the <DEVICE_NAME>?

The short answer is: in cbuild.yml yes, in CMakeLists.txt not yet.

However to proper advise on this matter we would need more info about your design:
1. Are the configuration.data and specs files being referenced in a component (e.g. Startup) that must be always selected?
2. Should the user be allowed to modify these files? Or are they immutable for a given pack version?

We can extend the information that is generated for each project ([cbuild.yml](https://open-cmsis-pack.github.io/cmsis-toolbox/YML-CBuild-Format/#cbuildyml)) and instruct [cbuild2cmake](https://github.com/Open-CMSIS-Pack/cbuild2cmake) to ultimately make them appear in the generated CMakeLists.txt, but we need to decide together the best approach since it deviates a bit from other supported toolchains.
We would appreciate to discuss this with you in the next Open-CMSIS-Pack meeting on Tuesday Jan 13th.

ashish-mahanth:
> Are the configuration.data and specs files being referenced in a component (e.g. Startup) that must be always selected?

Yes, specs and configuration.data files must always be passed to the compiler.

> Should the user be allowed to modify these files? Or are they immutable for a given pack version?

No, the user isn't allowed to modify these files.
 
Another option that I am thinking of is to extract the device name from "CONTEXT" variable e.g:"Example.Debug+PIC32CM6408PL10048"). Any thoughts on this ?

> We would appreciate to discuss this with you in the next Open-CMSIS-Pack meeting on Tuesday Jan 13th.

Sure, we will join the call this week.

jkrech:
@swaroopbekal-mchp no need for you to do anything, I will ask @brondani to take care of that.

jkrech (commented):
LGTM

I guess we should add tests potentially in another PR. 
Where can we download the XC32 toolchain from within CI?

swaroopbekal-mchp:
We found an issue in our internal testing. We need to make a minor update to this PR to correctly adjust the DFP path that we are passing to the compiler for some older DFPs. Let's not merge this right now.

PR #2340, opened 2026-01-29
Enable Star-MC3 for AC6 and GCC via Cortex-M52 compatibility

JonatanAntoni:
Neither AC6 nor GCC have native support for ArmChina's Star-MC3 core. But as MC3 is very close to Cortex-M52 we should be fine falling back to M52.


brondani (approved):
LGTM

PR #2430, opened 2026-04-08
Refine discoverLayer errMsg

jthuangarm:
To address: https://github.com/Open-CMSIS-Pack/vscode-cmsis-solution/issues/119
Updates the error message shown when no compatible software layer is found.

**User-facing improvements:**

* Updated the error message in `ProjMgrRpcServer.cpp` to clarify that no compatible software layer was found in the installed packs and to instruct users to install additional packs before restarting the "Create Solution" flow.

**Testing updates:**

* Updated the corresponding test in `ProjMgrRpcTests.cpp` to expect the new, more descriptive error message.

### DIFF
--- a/tools/projmgr/src/ProjMgrRpcServer.cpp
+++ b/tools/projmgr/src/ProjMgrRpcServer.cpp
@@ -836,8 +836,7 @@ RpcArgs::DiscoverLayersInfo RpcHandler::DiscoverLayers(const string& solution, c
   StrVec layers;
   StrSet fails;
   if(!m_worker.ListLayers(layers, "", fails) || !m_worker.ElaborateVariablesConfigurations()) {
-    result.message = "No compatible software layer was found in the installed packs. "
-                     "Install additional packs containing suitable layers before restarting the \"Create Solution\" flow.";
+    result.message = "No compatible software layer was found in the installed packs.\nInstall additional packs containing suitable layers before restarting the 'Create Solution' flow.";
     return result;
   } else {
     // retrieve valid configurations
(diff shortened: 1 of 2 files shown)


jthuangarm:
@brondani After discussing with @jkrech , we have come up with the following proposed refinement. Do you have any suggestions for the message?

tools/projmgr/src/ProjMgrRpcServer.cpp
```diff
@@ -836,7 +836,8 @@ RpcArgs::DiscoverLayersInfo RpcHandler::DiscoverLayers(const string& solution, c
   StrVec layers;
   StrSet fails;
   if(!m_worker.ListLayers(layers, "", fails) || !m_worker.ElaborateVariablesConfigurations()) {
-    result.message = "No compatible software layer found. Review required connections of the project";
+    result.message = "No compatible software layer was found in the installed packs. "
+                     "Install additional packs containing suitable layers before restarting the \"Create Solution\" flow.";
```
brondani:
```suggestion
    result.message = "No compatible software layer was found in the installed packs.\nInstall additional packs containing suitable layers before restarting the \'Create Solution\' flow.";
```

tools/projmgr/test/src/ProjMgrRpcTests.cpp
```diff
@@ -1498,7 +1498,8 @@ TEST_F(ProjMgrRpcTests, RpcDiscoverLayers) {
     json({{ "solution", csolutionPath }, { "activeTarget", "" }}));
   responses = RunRpcMethods(requests);
   EXPECT_FALSE(responses[0]["result"]["success"]);
-  EXPECT_EQ(responses[0]["result"]["message"], "No compatible software layer found. Review required connections of the project");
+  EXPECT_EQ(responses[0]["result"]["message"], "No compatible software layer was found in the installed packs. "
+    "Install additional packs containing suitable layers before restarting the \"Create Solution\" flow.");
```
brondani:
```suggestion
  EXPECT_EQ(responses[0]["result"]["message"], "No compatible software layer was found in the installed packs.\nInstall additional packs containing suitable layers before restarting the \'Create Solution\' flow.");
```

brondani (commented):
I agree with the proposed refinement.
For consistency I suggest to use single quotes instead of double quotes, and I believe a line break `\n` is needed for correct visualization of the message in two lines.

brondani (approved):

PR #2432, opened 2026-04-09
[projmgr] Clarify scope of debug, quiet, and verbose CLI options

jthuangarm:
This pull request updates the descriptions for the `--debug`, and also the `--quiet` and `--verbose` (for consistency) command-line options in `ProjMgr.cpp` to clarify that these flags specifically control messages from the CMSIS build tools. There are no changes to functionality.

### DIFF
--- a/tools/projmgr/src/ProjMgr.cpp
+++ b/tools/projmgr/src/ProjMgr.cpp
@@ -44,7 +44,7 @@ Commands:\n\
 Options:\n\
   -a, --active arg              Select active target-set: <target-type>[@<set>]\n\
   -c, --context arg [...]       Input context names [<project-name>][.<build-type>][+<target-type>]\n\
-  -d, --debug                   Enable debug messages\n\
+  -d, --debug                   Enable debug messages of the cmsis build tools\n\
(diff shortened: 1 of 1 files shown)


brondani:
The projmgr command-line options refer only to `csolution`, I believe that the text "of the cmsis build tools" is misleading here. This should be rather applied to `cbuild` help text as long as [cbuild](https://github.com/Open-CMSIS-Pack/cbuild) is actually the cmsis build orchestrator.
@jkrech Do you have any other suggestion about the `csolution` help text or should this PR be closed without changes?

jkrech:
I agree with @brondani. `csolution` is not invoking other tools subsequently, therefore all command line options refer the tool itself.
In `cbuild` this is different, it explicitly invokes subsequent tools and a command line option could be "targeted" to only one of these tools.
E.g.:
```
-t, --target arg         Optional CMake target name
```
This option is not targeted to cpackget, cbuild2cmake, ... but only to cmake

I suggest to close this PR.

jthuangarm:
Thanks for the clarification and suggestion. Will make the changes under `cbuild`.

Issue #2216, opened 2025-09-18
Library build should exclude component files


ReinhardKeil:
**The Problem To Be Solved**

When a library is generated in uVision, files from software components are excluded.

See https://developer.arm.com/documentation/101407/0543/Creating-Applications/Software-Components/Options-for-Component?lang=en

This feature is currently not available in CMSIS-Toolbox which results in https://github.com/ARM-software/vscode-cmsis-csolution/issues/334

**Suggest A Solution**

- Add a yml node `build-scope: {include | exclude}` to [`component:`](https://open-cmsis-pack.github.io/cmsis-toolbox/YML-Input-Format/#components) with this default: 
   - `build-scope: include` for executable build
   - `build-scope: exclude` for library build

This would exclude source code for software components in library build, and include it in a standard build. It also gives the same level of control as uVision did provide.

jkrech:
Validating: csolution 2.13.0+p48-gf85c8499 (C) 2022-2026 Arm Ltd. and Contributors

when setting
```
  output:
    type:
      - lib
```
source files from components are no longer listed in cbuild.yml except for components with `build-scope: include`.

When creating e.g an `elf` file as output,

```
  output:
    type:
      - elf
``` 

by default all source files from components are included in the build, except when specifying a component `build-scope: exclude`.

New node `build-scope` is mentioned in the documentation in the table for [components](https://open-cmsis-pack.github.io/cmsis-toolbox/YML-Input-Format/#components) but no specific section / example to link to from the schema.

ReinhardKeil:
added to documentation (https://open-cmsis-pack.github.io/cmsis-toolbox/build-overview/#generate-library)

PR #2436, opened 2026-04-14
Add `build-scope` to include/exclude component files

jthuangarm:
To address: https://github.com/Open-CMSIS-Pack/devtools/issues/2216

Introduces support for a new `build-scope` attribute for components to control whether component source files are included in different build types (library vs. executable). By default, component source files are:
* included in executable builds
* excluded from library builds

**Changes**

* Added `build-scope` attribute (`include | exclude`) to component definitions.
* Updated schema (common.schema.json) and YAML parsing to support and validate the new attribute.
* Updated build generation to include/exclude component source files, based on `build-scope` and build type.
* Added helper method `IsSourceCategory` to identify compatible source files.

### DIFF
--- a/tools/projmgr/test/data/TestSolution/StandardLibrary/library_componentBuildScope/library_componentBuildScope.cproject.yml
+++ b/tools/projmgr/test/data/TestSolution/StandardLibrary/library_componentBuildScope/library_componentBuildScope.cproject.yml
@@ -1,0 +1,13 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/Open-CMSIS-Pack/devtools/main/tools/projmgr/schemas/cproject.schema.json
+
+project:
+
+  output:
+    type: lib
+
+  components:
+    - component: CORE
+      build-scope: exclude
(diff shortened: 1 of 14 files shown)


tools/projmgr/test/data/TestSolution/StandardLibrary/library_componentBuildScope/library_componentBuildScope.cproject.yml
```diff
@@ -0,0 +1,13 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/Open-CMSIS-Pack/devtools/main/tools/projmgr/schemas/cproject.schema.json
+
+project:
+
+  output:
+    type: lib
+
+  components:
+    - component: CORE
+      build-scope: exclude
```
brondani: Another context with `elf` output should be used to test the remaining "exclude" behavior which it seems it's not yet covered.
I believe the `CORE` component does not have source files, can you please check? In such case a different component should be used.

brondani (approved):
LGTM

Issue #2424, opened 2026-04-03
Add `default` attribute to the device memory in .cbuild-run.yml


TeoMahnic:
**Describe The Problem To Be Solved**
The CMSIS-Pack specification defines a `default` attribute for memory elements, indicating a general purpose memory region that does not require special handling. This attribute is currently not exposed in the `.cbuild-run.yml` under the `system-resources: memory` node.

**Use Case**
RTT control block search in pyOCD requires identifying default memory regions, which is currently not possible without this attribute.

jkrech:
validating nightly 2026-04-15: csolution 2.13.0+p48-gf85c8499 (C) 2022-2026 Arm Ltd. and Contributors

Using ModelNova example:
SDS+AppKit-E8-U85.cbuild-run.yml
```
 system-resources:
    memory:
      - name: MRAM
        access: rx
        start: 0x80000000
        size: 0x00580000
        default: true
        from-pack: AlifSemiconductor::Ensemble@2.1.0
...
     - name: SRAM1
        access: rwx
        start: 0x08000000
        size: 0x00400000
        default: false
        from-pack: AlifSemiconductor::Ensemble@2.1.0
```

It now lists the memory node `default:` 
This is generated from pdsc memory information for the selected device:
`<memory name="MRAM" start="0x80000000" size="0x00580000" access="rx" default="1"/>`

`<memory name="SRAM1" start="0x02000000" size="0x00400000" access="rwx" default="0"/>`

PR #2438, opened 2026-04-14
Add `default` attribute to `memory` in cbuild-run.yml

jthuangarm:
To address https://github.com/Open-CMSIS-Pack/devtools/issues/2424

This pull request introduces support for marking memory regions as `default` in the project manager tool. In cbuild-run.yml `default` is located right before the `from-pack:` property under `memory`.

* Updated the JSON schema (`common.schema.json`) to define the new `default` property for memory objects.
* Modified `ProjMgrRunDebug.cpp` to read the `default` attribute when collecting settings and to write it when generating YAML output.


brondani (approved):
LGTM

PR #2457, opened 2026-04-24
[projmgr] Added debugger systemview node support

soumeh01:
## Changes
- Schema extension for https://open-cmsis-pack.github.io/cmsis-toolbox/pyOCD-Debugger/#systemview

## Checklist


tools/projmgr/schemas/common.schema.json
```diff
@@ -2322,6 +2323,17 @@
...
       "additionalProperties": false
     },
+    "SystemViewOptionType": {
+      "title": "systemview:\nDocumentation: https://open-cmsis-pack.github.io/cmsis-toolbox/pyOCD-Debugger/#systemview",
+      "description": "Configuration for SystemView features.",
+      "type": "object",
+      "properties": {
+        "file":             { "type": "string", "description": "Path to the SystemView capture file." },
+        "auto-start":       { "type": "boolean", "description": "Send SystemView start command automatically.", "default": true },
+        "auto-stop":        { "type": "boolean", "description": "Send SystemView stop command automatically.", "default": true }
```
TeoMahnic: These are the only two properties with a `default` value. Does this have any functional impact?

tools/projmgr/schemas/common.schema.json
```diff
@@ -2322,6 +2323,17 @@
       },
       "additionalProperties": false
     },
+    "SystemViewOptionType": {
+      "title": "systemview:\nDocumentation: https://open-cmsis-pack.github.io/cmsis-toolbox/pyOCD-Debugger/#systemview",
+      "description": "Configuration for SystemView features.",
+      "type": "object",
+      "properties": {
+        "file":             { "type": "string", "description": "Path to the SystemView capture file." },
```
brondani, line 2331: `file` is a path, and properly handling paths in csolution requires it to be relative to its parent file, i.e. *.csolution.yml and *.cbuild-run.yml, which are by default located in separate directories.
For this reason, simply adding this to the schema and relying on the one-to-one copy is not sufficient; for consistency, its relative value needs to be processed and recalculated. Support for access sequences should also be considered.
@TeoMahnic Is this already being used in the debugger extension? If yes I am afraid it is not consistent.

TeoMahnic (approved):
LGTM

tools/projmgr/schemas/common.schema.json
TeoMahnic: @brondani, yes this is already processed in pyOCD. Currently the `file:` node value from `*.cbuild-run.yml` is taken as is, but it would make sense that the path is always relative to the `*.cbuild-run.yml`.

brondani: I do not recommend making exceptions to the rule of relative paths, as this would introduce inconsistency.
If you prefer to avoid processing paths, I suggest changing the `file` description and documentation to treat it as a “filename” rather than a “path”, with its base directory fixed to the `out` directory (along side *.cbuild-run.yml).

tools/projmgr/schemas/common.schema.json
brondani: Indeed we don't set default values for properties in the schema, I suggest to remove it.

soumeh01: [Removing default values](https://github.com/Open-CMSIS-Pack/devtools/pull/2457/commits/2711a5e565ae67bfb0f851f93aa92000a6928a4d)

tools/projmgr/schemas/common.schema.json
jkrech: I do not think that fixing the file location of the `out` directory gives us the consistency and flexibility, expecially if users want to separate build output from debug output. 
Therefore csolution must actively convert the relative file path during the generation of the `cbuild-run.yml` file

Note that the  "TelnetOptionType": also has a file node that requires this processing.

brondani (approved):
LGTM

PR #2514, opened 2026-07-10
Add debugger `device-settings` support to `cbuild-run` generation

brondani:
## Fixes
- https://github.com/Open-CMSIS-Pack/cmsis-toolbox/issues/636

## Changes
- Added `device-settings` to debugger parsing, schema validation, context processing, and `cbuild-run.yml` output.
- Added fallback extraction of debugger device settings from `.dbgconf` variable assignments when explicit `device-settings` are not provided.
- Added `ParseDbgconfFile` utility coverage and updated run-debug reference outputs/test data.

## Risk / Limitations
- `.dbgconf` parsing is limited to simple semicolon-terminated assignments after stripping C-style comments.

## Checklist

### DIFF
--- a/tools/projmgr/src/ProjMgrUtils.cpp
+++ b/tools/projmgr/src/ProjMgrUtils.cpp
@@ -42,7 +42,7 @@ static string Trim(const string& value) {
 static string StripCComments(const string& input) {
   string output;
   bool inLineComment = false;
-  bool inBlockComment = false;
+  size_t blockCommentDepth = 0;
   size_t index = 0;
   while (index < input.length()) {
     const char current = input[index];
@@ -54,9 +54,12 @@ static string StripCComments(const string& input) {
         output += current;
       }
       ++index;
-    } else if (inBlockComment) {
-      if (current == '*' && next == '/') {
-        inBlockComment = false;
+    } else if (blockCommentDepth > 0) {
+      if (current == '/' && next == '*') {
+        ++blockCommentDepth;
+        index += 2;
+      } else if (current == '*' && next == '/') {
+        --blockCommentDepth;
         index += 2;
       } else {
         if (current == '\n' || current == '\r') {
@@ -68,7 +71,7 @@ static string StripCComments(const string& input) {
       inLineComment = true;
       index += 2;
     } else if (current == '/' && next == '*') {
-      inBlockComment = true;
+      blockCommentDepth = 1;
       index += 2;
     } else {
       output += current;
--- a/tools/projmgr/src/ProjMgrUtils.cpp
+++ b/tools/projmgr/src/ProjMgrUtils.cpp
@@ -13,6 +13,7 @@
 #include "CrossPlatform.h"
 #include "CrossPlatformUtils.h"
 
+#include <cctype>
 #include <regex>
 #include <sstream>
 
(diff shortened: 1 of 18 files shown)


brondani:
@grasci-arm Please review this PR

grasci-arm:
Thanks for the ping. I’ll review the PR and get back with comments if I find anything.

grasci-arm (commented):
The new device-settings plumbing is broadly coherent, but I found one serious regression risk: dbgconf parsing now silently ignores any final assignment lacking a semicolon and will drop the entire file’s last setting, which can break generated debugger behavior for valid-but-minimal dbgconf content.

grasci-arm:
Thanks for the ping — I’ve reviewed the PR and the new `device-settings` flow looks consistent across parsing, validation, context processing, and `cbuild-run.yml` generation.

A couple of things stand out as good to keep in mind:
- the fallback extraction from `.dbgconf` is limited to simple assignments, so the documented limitation is important
- the updated tests/reference outputs seem to cover the intended paths well

Overall, the change looks good to me.

grasci-arm:
Thanks for the ping — I’ll review this PR.

tools/projmgr/src/ProjMgrUtils.cpp
```diff
@@ -22,6 +23,61 @@ static constexpr const char* HIGHER_OR_EQUAL_OPERATOR = ">=";
...
+
+// Remove surrounding whitespace without changing the assignment value content.
+static string Trim(const string& value) {
+  const auto first = find_if_not(value.begin(), value.end(), [](unsigned char c) { return isspace(c); });
+  const auto last = find_if_not(value.rbegin(), value.rend(), [](unsigned char c) { return isspace(c); }).base();
+  return first < last ? string(first, last) : string();
+}
+
+// Strip C and C++ comments while preserving line breaks that separate statements.
+static string StripCComments(const string& input) {
```
edriouk, line 42: Could we have a dedicated unit test for StripCComments? 
In particular we need to test nested comments and combinations of block and in-line comments

edriouk (changes requested):
Some additional unit tests are desirable

tools/projmgr/src/ProjMgrUtils.cpp
brondani: `StripCComments` is file-local and only reached through `ParseDbgconfFile`, so the useful test surface is the existing `ProjMgrUtilsUnitTests.ParseDbgconfFile`. I’m going to extend it to cover nested/block/in-line combinations.

edriouk: Why it is hidden? The method does not change any static or global variable.
I would suggest making `Trim`, `StripCSomments` and `IsDbgconfVariableName` `public static` methods of `ProjMgrUtils `class

brondani: `Trim`, `StripCComments` and `IsDbgconfVariableName` are implementation details of the public `ProjMgrUtils::ParseDbgconfFile`. They are not used anywhere else. Making them public static would mostly add API surface without a caller-driven need. It also freezes names/behavior that are currently free to change as parser internals. For example, `Trim` is generic-sounding but currently exists only to support dbgconf parsing; exposing it through `ProjMgrUtils` would make it look like a general utility contract.

edriouk (approved):
LGTM

PR #2528, opened 2026-07-20
Added Infineon_Device_Configurator and InfineonMcuConfigWizard

OttJulia:
Added Infineon_Device_Configurator and InfineonMcuConfigWizard (MOTIX and LITIX devices)

## Fixes
- none

## Changes
- Added new generators

## Checklist


jkrech (changes requested):
a) There already exists a global generator with ID `InfineonDeviceConfigurator`. You are adding a global generator ID `Infineon_Device_Configurator`  using `_` this can be easily overlooked and can cause user confusion.
The download-url is identical: `https://softwaretools.infineon.com/tools/com.ifx.tb.tool.deviceconfigurator`
However the `run:` node and the `path:` are different.
Is there a good reason why this cannot be unified?

b) The "run:" node specified a `*.bat` file which only executes on windows, specify a launch script without extension and then provide launch scripts with different launch scripts for each platform
- darwin-arm-64 (bash)
- linux-amd64 (bash)
- linux-arm64 (bash)
- windows-amd64 (bat)
- windows-arm64 (bat)
even if the invocation just prints: `Platform is not supported for Infineon Device Configurator!` and `exit 1`.
The generator much not silently fail and leave the user puzzled.

c) The launch scripts need to be added via a PR in this repository: 
https://github.com/Open-CMSIS-Pack/generator-bridge/tree/main/scripts 
The existing Infineon script are here:
https://github.com/Open-CMSIS-Pack/generator-bridge/tree/main/scripts/Infineon_Device_Configurator

d) For the InfineonMcuConfigWizard scripts need to be added via a PR as well:
https://github.com/Open-CMSIS-Pack/generator-bridge/tree/main/scripts/InfineonMcuConfigWizard

OttJulia:
a) I see where the confusion comes from, especially because InfineonDeviceConfigurator uses a folder named Infineon_Device_Configurator in the bridge repo. 
The reason for added the same tool as a new generator is that the path arguments are different. Two independent teams selected different file structures for their examples, and now they are like that and won't be changed. I was thinking about adding the differentiation based on device name to the batch, but that seemed not very safe for the future.
I could rename it to InfineonDeviceConfigurator_MotixLitix and integrate the device families in the generator name - what do you think?

b/c/d) Thanks, I will update the run element and add the scripts once the name is clarified

jkrech:
Thanks for sharing the background.
Your proposal for the global generator ID sounds sensible as it does make the ID unique. 
I am unable to judge how familiar your end users are with MotixLitix but I assume this is fine.
The only concern is the length of the ID 
Maybe `InfineonDevConf_MotixLitix` works for you as well.

jkrech:
@OttJulia, could you please create the PR for the launch scripts you are referencing
`https://github.com/Open-CMSIS-Pack/generator-bridge/tree/main/scripts`
We would merge both PRs for the same nightly build of the CMSIS-Toolbox.

OttJulia:
Created a PR for the generator files: https://github.com/Open-CMSIS-Pack/generator-bridge/pull/355

brondani (approved):
To package the scripts correctly in CMSIS-Toolbox, the following workflows and integration test also need to be extended in the [cmsis-toolbox](https://github.com/Open-CMSIS-Pack/cmsis-toolbox) repository:

- [E2E workflow](https://github.com/Open-CMSIS-Pack/cmsis-toolbox/blob/25f48aeb5bffa1814276127154e1651751c277c2/.github/workflows/shared-e2e.yml#L302)
- [Toolbox workflow](https://github.com/Open-CMSIS-Pack/cmsis-toolbox/blob/25f48aeb5bffa1814276127154e1651751c277c2/.github/workflows/toolbox.yml#L293)
- [Packaging Test](https://github.com/Open-CMSIS-Pack/cmsis-toolbox/blob/25f48aeb5bffa1814276127154e1651751c277c2/test/packaging/tests.py#L16-L19)

jkrech (approved):
LGTM

PR #2559, opened 2026-08-14
Add possibility to use piped attributes for comparison

edriouk:
## Fixes

This address possibility to use aliases for target attributes, e.g. for a compiler.
Considering csolution.yml:

```
compiler: XC
alias: GCC
```

This change will allow to set RteTarget` attribute Tcompiler to "XC|GCC" effectively to accept all components for GCC and XC compilers

## Changes

- Add` WildCard::MatchAny()` method to compare alternatives from both arguments "A|B|C"
- Use `MatchAny` when evaluating generic target attributes in `RteConditionExpression`.
- Ad corresponding tests

## Checklist


brondani (approved):
LGTM

PR #2581, opened 2026-08-31
[ctrace] Align trace-run metadata with specification

thorstendb-ARM:
## Fixes
- Aligns ctrace with the current experimental CMSIS-Toolbox trace-run specification.
- https://open-cmsis-pack.github.io/cmsis-toolbox/Experimental-Features/#ctrace-utility

## Changes

- Read DWT metadata from the current `address`, `size`, and `data-type` fields.
- Support the specified `unsigned`, `signed`, and `float` data types.
- Use referenced `ctrace-setup.data.size` when reference-level `size` is absent.
- Distinguish DWT `data#<index>` references from instruction-control references so that only data references create decoded data sources.
- Ignore unknown and unconsumed YAML fields while retaining validation for fields used by ctrace.
- Update output planning, CTF metadata, fixtures, and tests to use the current field names.
- Remove compatibility handling for the previous field names because ctrace has not been released yet.

## Checklist

### DIFF
--- a/tools/ctrace/src/output/ctf/CtfSchema.h
+++ b/tools/ctrace/src/output/ctf/CtfSchema.h
@@ -81,8 +81,8 @@ inline constexpr std::array<ValueVariant, 7U> ValueVariants{{
 }};
 
 inline constexpr std::string_view ValueTypeRequirements =
-    "supported data.symbol-type values are 'unsigned int', 'signed int', and 'float'; "
-    "data.symbol-size must be 1, 2, or 4, and float requires size 4";
+    "supported data-type values are 'unsigned', 'signed', and 'float'; "
+    "size must be 1, 2, or 4, and float requires size 4";
 
 /** @brief Returns the schema descriptor for a value tag. */
 constexpr const ValueVariant& valueVariant(ValueTag tag)
@@ -97,8 +97,8 @@ constexpr const ValueVariant* valueVariantForTraceRunType(const std::string_view
     return byteSize == 4U ? &valueVariant(ValueTag::Float32) : nullptr;
   }
 
-  const auto signedInteger = typeName == "signed int";
-  if (!signedInteger && typeName != "unsigned int") {
+  const auto signedInteger = typeName == "signed";
+  if (!signedInteger && typeName != "unsigned") {
     return nullptr;
   }
   if (byteSize == 1U) {
--- a/tools/ctrace/README.md
+++ b/tools/ctrace/README.md
@@ -2,13 +2,7 @@
 
 `ctrace` converts CMSIS Cortex-M trace captures into outputs that can be inspected or processed by other tools. It
 combines a `<solution-set>.ctrace-run.yml` description with matching raw trace files from one trace directory.
-
-The current implementation decodes SWO/ITM and DWT data. Trace Bus (`*.TB.raw`) files are discovered and reported,
-but deliberately skipped until that decoder is implemented.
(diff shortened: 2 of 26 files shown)


soumeh01 (commented):
`svdconv-cov` shows a -0.44% coverage change although this PR does not modify SVDConv.

The SVDConv workflow is path-filtered and is not run for these ctrace only changes. The `svdconv-cov` Codecov flag has `carryforward: true`, so this appears to be a carried-forward/stale coverage comparison rather than a coverage regression introduced by this PR.

`carryforward: true` is useful for our use case as the repository is large. SVDConv wasn't tested this time, so It will carry forward its previous coverage. Otherwise it will complain about the 0 coverage/missing coverage.

jreineckearm:
> carryforward: true is useful for our use case as the repository is large. SVDConv wasn't tested this time, so It will carry forward its previous coverage. Otherwise it will complain about the 0 coverage/missing coverage.

@soumeh01 , does the failing coverage check gate this PR once a reviewer approved?

tools/ctrace/src/output/ctf/CtfSchema.h
```diff
@@ -81,8 +81,8 @@ inline constexpr std::array<ValueVariant, 7U> ValueVariants{{
 }};
 
 inline constexpr std::string_view ValueTypeRequirements =
-    "supported data.symbol-type values are 'unsigned int', 'signed int', and 'float'; "
-    "data.symbol-size must be 1, 2, or 4, and float requires size 4";
+    "supported data-type values are 'unsigned', 'signed', and 'float'; "
+    "size must be 1, 2, or 4, and float requires size 4";
```
jreineckearm, line 85: How would a full message with this look like? Does it make clear that this is only CTF generation?

soumeh01 (dismissed):

soumeh01:
> > carryforward: true is useful for our use case as the repository is large. SVDConv wasn't tested this time, so It will carry forward its previous coverage. Otherwise it will complain about the 0 coverage/missing coverage.
> 
> @soumeh01 , does the failing coverage check gate this PR once a reviewer approved?

Seems like it doesn't I just approved and it allows me to merge it.

jreineckearm (commented):
Just one question, will give it a try next

jreineckearm (dismissed):
Tested, still works as expected with the new nodes written by pyTS.
Good to get merged (I can't, I don't have the rights).

soumeh01:
@thorstendb-ARM All comments must be resolved to merge this.

tools/ctrace/src/output/ctf/CtfSchema.h
thorstendb-ARM: Would look like this:
```
[info] selected trace-run configuration: solutionSet=InvalidDataType, path=.../InvalidDataType.ctrace-run.yml, references=1, setups=1
[error] CTF output cannot use ctrace-run data-type 'boolean'; supported data-type values are 'unsigned', 'signed', and 'float'; size must be 1, 2, or 4, and float requires size 4: backend=ctf, channel=DWT0, stream=0, config=.../InvalidDataType.ctrace-run.yml, dataType=boolean
```

jreineckearm: Great thanks for confirming. This gives enough hint it's about the CTF output where we have those restrictions.

jreineckearm (approved):

PR #2587, opened 2026-09-01
Ctrace exception return

thorstendb-ARM:
## Fixes

- Correctly represents exception return transitions in generated CTF traces.

## Changes

- Preserve `entered`, `exited`, and `returned` exception actions in CTF output.
- Distinguish trace-originated events from synthetic lane-closing events.
- Add a dedicated `Return` lane to the Trace Compass analysis.
- Order exception lanes as `ThreadMode`, `Return`, followed by individual exceptions.
- Reset active exception state on overflow or data loss, leaving unknown intervals as gaps.
- Extend unit and integration tests for the new event semantics and visualization.

## Validation

- Built `ctrace`, `CtraceUnitTests`, and `CtraceIntegTests`.
- All six ctrace test targets pass.
- Converted a real SWO exception trace with the current build.
- Validated the generated CTF using Babeltrace 2.
- Verified the state-system result in Trace Compass Trace Server.

## Checklist

### DIFF
--- a/tools/ctrace/src/output/ctf/CtfEncoder.cpp
+++ b/tools/ctrace/src/output/ctf/CtfEncoder.cpp
@@ -379,9 +379,13 @@ void CtfEncoder::writeExceptionEvent(std::uint8_t traceBusId, const ExceptionTra
 void CtfEncoder::emitExceptionRecord(std::uint8_t traceBusId, std::uint32_t number,
                                      CtfExceptionLaneTracker::RecordAction action)
 {
+  const auto semanticAction =
+      action == CtfExceptionLaneTracker::RecordAction::Enter
+          ? ExceptionAction::Entered
+          : action == CtfExceptionLaneTracker::RecordAction::Exit ? ExceptionAction::Exited : ExceptionAction::Returned;
   TraceEvent selectionEvent{ExceptionTraceEvent{
       number,
-      action == CtfExceptionLaneTracker::RecordAction::Enter ? ExceptionAction::Entered : ExceptionAction::Exited,
+      semanticAction,
   }};
   selectionEvent.traceBusId = traceBusId;
   if (!traceEventSelectedForOutput(selectionEvent, m_config.selection)) {
@@ -389,9 +393,11 @@ void CtfEncoder::emitExceptionRecord(std::uint8_t traceBusId, std::uint32_t numb
   }
   constexpr auto payloadSize = 2U + 1U + 2U;
   const auto eventTimestamp = allocateEventTimestamp(traceBusId);
-  const auto encodedAction =
-      CtfSchema::value(action == CtfExceptionLaneTracker::RecordAction::Enter ? CtfSchema::ExceptionAction::Entered
-                                                                              : CtfSchema::ExceptionAction::Exited);
+  const auto encodedAction = CtfSchema::value(
+      action == CtfExceptionLaneTracker::RecordAction::Enter
+          ? CtfSchema::ExceptionAction::Entered
+          : action == CtfExceptionLaneTracker::RecordAction::Exit ? CtfSchema::ExceptionAction::Exited
+                                                                  : CtfSchema::ExceptionAction::Returned);
   m_stream.writeRecord(CtfSchema::value(CtfSchema::EventId::Exception), eventTimestamp, traceBusId, payloadSize,
                        [&](CtfStreamWriter::Record& record) {
                          record.writeU16(static_cast<std::uint16_t>(number & 0xffffU));
--- a/tools/ctrace/src/output/ctf/CtfEncoder.cpp
+++ b/tools/ctrace/src/output/ctf/CtfEncoder.cpp
@@ -346,6 +346,18 @@ void CtfEncoder::writeGlobalTimestampEvent(const TraceEvent& event, const Global
 
 void CtfEncoder::writeTraceStatusEvent(std::uint8_t reason, std::uint8_t traceBusId, bool emitEvent)
 {
+  if (reason == CtfSchema::value(CtfSchema::TraceStatusReason::Overflow) ||
+      reason == CtfSchema::value(CtfSchema::TraceStatusReason::DataLoss)) {
+    const auto lane = m_exceptionLanes.find(traceBusId);
+    if (lane != m_exceptionLanes.end()) {
+      lane->second.resetForDiscontinuity(
+          [this, traceBusId](std::uint32_t number, CtfExceptionLaneTracker::RecordAction action,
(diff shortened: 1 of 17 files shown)


tools/ctrace/src/output/ctf/CtfEncoder.cpp
```diff
@@ -354,48 +366,49 @@ void CtfEncoder::writeTraceStatusEvent(std::uint8_t reason, std::uint8_t traceBu
...
       number,
-      action == CtfExceptionLaneTracker::RecordAction::Enter ? ExceptionAction::Entered : ExceptionAction::Exited,
+      semanticAction,
   }};
   selectionEvent.traceBusId = traceBusId;
   if (!traceEventSelectedForOutput(selectionEvent, m_config.selection)) {
     return;
   }
-  constexpr auto payloadSize = 2U + 1U + 2U;
+  constexpr auto payloadSize = 2U + 1U + 2U + 1U;
```
JonatanAntoni, line 396: Algorithmic question: Why do we need to pre-calculate the size here? Could we deduce the size from the writes happening later? Or do we need it beforehand to pre-allocate buffers with the right size?

tools/ctrace/src/output/ctf/CtfEncoder.cpp
```diff
@@ -354,48 +366,49 @@ void CtfEncoder::writeTraceStatusEvent(std::uint8_t reason, std::uint8_t traceBu
...
+      action == CtfExceptionLaneTracker::RecordAction::Enter
+          ? CtfSchema::ExceptionAction::Entered
+          : action == CtfExceptionLaneTracker::RecordAction::Exit ? CtfSchema::ExceptionAction::Exited
+                                                                  : CtfSchema::ExceptionAction::Returned);
+  const auto encodedOrigin = CtfSchema::value(origin == CtfExceptionLaneTracker::RecordOrigin::Trace
+                                                  ? CtfSchema::ExceptionOrigin::Trace
+                                                  : CtfSchema::ExceptionOrigin::Synthetic);
   m_stream.writeRecord(CtfSchema::value(CtfSchema::EventId::Exception), eventTimestamp, traceBusId, payloadSize,
                        [&](CtfStreamWriter::Record& record) {
                          record.writeU16(static_cast<std::uint16_t>(number & 0xffffU));
```
JonatanAntoni: We may use standard macros instead of hard-coded constants:
```suggestion
                         record.writeU16(static_cast<std::uint16_t>(number & UINT16_MAX));
```

JonatanAntoni: Instead of repeating this kind of conversion all over the place, why not add this into `record.writeU16()` and make the function accept uint32_t (or others)? E.g., by simple a overload or using a template...

tools/ctrace/src/output/ctf/CtfEncoder.h
```diff
@@ -80,7 +80,8 @@ class CtfEncoder final {
   /** @brief Applies one exception transition to its CTF lane state. */
   void writeExceptionEvent(std::uint8_t traceBusId, const ExceptionTraceEvent& exception);
   /** @brief Emits one concrete exception lane record. */
-  void emitExceptionRecord(std::uint8_t traceBusId, std::uint32_t number, CtfExceptionLaneTracker::RecordAction action);
+  void emitExceptionRecord(std::uint8_t traceBusId, std::uint32_t number, CtfExceptionLaneTracker::RecordAction action,
```
JonatanAntoni: The exception record actually asks for `uint16_t number`. Why does the function accept a larger range?

tools/ctrace/test/unit/src/output/ctf/CtfEncoderTests.cpp
```diff
@@ -328,19 +328,20 @@ TEST(CtraceUnitTests, testCtfEncoderStreamSelectionKeepsStartAndResyncContext)
...
                                CtfSchema::value(CtfSchema::TraceStatusReason::TraceStart),
                                CtfSchema::value(CtfSchema::TraceStatusReason::Resync),
                            }));
   EXPECT_EQ(exceptionRecords, (std::vector<std::pair<std::uint64_t, std::string>>({
-                                  {0U, "0:0"},
-                                  {10U, "0:1"},
-                                  {10U, "15:0"},
-                                  {20U, "15:1"},
-                                  {20U, "54:0"},
+                                  {0U, "0:0:1"},
```
JonatanAntoni: Knowledge about this format is spread across the test. What about capture the above logic into a helper function?

JonatanAntoni (dismissed):
LGTM, see some neat picking comments

tools/ctrace/test/unit/src/output/ctf/CtfMetadataWriterTests.cpp
```diff
@@ -90,6 +92,19 @@ TEST(CtraceUnitTests, testTraceCompassXmlUsesCurrentCtfEvents)
   for (const auto eventId : visualizedEvents) {
     EXPECT_NE(xml.find("eventName=\"" + std::string(CtfSchema::eventName(eventId)) + "\""), std::string::npos);
   }
+  EXPECT_NE(xml.find("value=\"returned\""), std::string::npos);
+  EXPECT_NE(xml.find("value=\"cmsis_exception_origin\""), std::string::npos);
+  EXPECT_NE(xml.find("value=\"trace\""), std::string::npos);
+  EXPECT_NE(xml.find("value=\"EXCEPTION_RETURN\""), std::string::npos);
+  EXPECT_NE(xml.find("value=\"Exception Return\""), std::string::npos);
+  const auto threadModeEntry = xml.find("path=\"EXCEPTION/Thread Mode\"");
+  const auto returnEntry = xml.find("path=\"EXCEPTION_RETURN/" "*\" displayText=\"true\"");
```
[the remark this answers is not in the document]
thorstendb-ARM: This split was intentional. Both literals are concatenated by the compiler, but splitting them prevents a literal /* sequence in the source. comment-parser, used by the Python copyright checker, incorrectly interprets /* inside the C++ string as the start of an unterminated block comment. This was a temporary workaround until the checker is fixed.

thorstendb-ARM:
Thanks, addressed all review comments:
- Simplified the concatenated XML string literal.
- Introduced ExceptionNumber as uint16_t and use it consistently, removing masks and casts from the CTF encoder.
- Kept writeU16(uint16_t) strict to prevent implicit truncation.
- Centralized CTF exception-record decoding and validation in CtfTestSupport.
- payloadSize is required before the callback so CtfStreamWriter can handle packet rollover and provide bounded record writes. I documented this requirement at the API.

JonatanAntoni (approved):

PR #2592, opened 2026-09-04
feat(ctrace): decode Armv8-M DWT match packets

thorstendb-ARM:
## Summary

Add support for Armv8-M Data Trace Match packets.

- Decode comparator-only match packets instead of reporting an unsupported address payload.
- Emit CSV `dwt` rows with the comparator in `source`; value, PC, and offset remain empty.
- Add a compact `DWT_MATCH` CTF event containing the comparator and trace-quality information.
- Visualize matches in Trace Compass as 1 µs `Something happened` pulses.
- Preserve pending DWT fragments before emitting a match.

## Tests

- Added unit tests for decoding, CSV, CTF, filtering, metadata, and XML.
- Added an 18-byte synthetic Armv8-M SWO fixture containing synchronization, match packets for comparators 0–3, and local timestamps.
- The fixture is generated from the architecture specification and was not captured from real hardware.
- All ctrace unit, integration, and executable tests pass.

## Checklist


jreineckearm (approved):
Changes look reasonable, not tested though due to the lack of test HW.

PR #2594, opened 2026-09-04
feat(ctrace): preserve DWT address fragment widths

thorstendb-ARM:
## Summary

Preserve the raw width and value of both Armv8-M DWT address-bearing packet types:

- **Data Trace Data Address** reports the memory address associated with a Data Address comparator match.
- **Data Trace PC Value** reports the instruction address associated with a comparator action, for example an Instruction Address match or a Data Address watch configured to emit its associated PC.

Both packet types can carry a 1-, 2-, or 4-byte low-order address fragment. The fragment is not an arithmetic difference.

## Changes

- accept 1-, 2-, and 4-byte Data Trace Data Address and Data Trace PC Value payloads
- retain each raw address fragment together with its encoded width
- write Data Trace Data Address values to the CSV `address` column
- write Data Trace PC Value values to the CSV `pc` column
- format both CSV values using their original width
- use shared address-fragment handling across decoder, model, CSV, and CTF output
- replace the former semantic `offset` identifiers with `address` throughout the DWT model and decoder
- encode optional PC and address fields in CTF as compact `none`/`u8`/`u16`/`u32` variants
- name the corresponding CTF fields `cmsis_dwt_pc*` and `cmsis_dwt_address*`
- keep DWT address values available as numeric scalars in Trace Compass
- add decoder, pipeline, CSV, CTF metadata, encoder, and Trace Compass XML coverage

Examples:

- 1 byte: `0x58`
- 2 bytes: `0x7858`
- 4 bytes: `0x20007858`

The terminology change applies to the decoded model and output formats. The ctrace configuration values `offset`, `offset+value`, and `PC+offset` remain unchanged.

## Packet distinction

A one-byte Data Trace Match payload with value `0x01` is recognized before Data Trace PC Value decoding. Data Trace Data Value packets and periodic PC sampling packets are separate formats and are unchanged by this PR.

This intentionally preserves the address fragment received from the trace stream. It does not reconstruct omitted upper address bits using `DWT_COMP<n>` or information from `ctrace-run.yml`.

## Validation

- Release unit and integration tests passed
- Debug/coverage unit and integration tests passed
- generated CTF metadata and streams were successfully read with Babeltrace
- the repository-wide copyright check has no missing notices; its only failure is the known `comment-parser` error caused by the pre-existing `EXCEPTION_RETURN/*` test string on `main`

## Checklist

### DIFF
--- a/tools/ctrace/src/decode/DwtPacketDecoder.cpp
+++ b/tools/ctrace/src/decode/DwtPacketDecoder.cpp
@@ -54,9 +54,14 @@ constexpr std::uint32_t kExceptionActionShift = 12U;
 constexpr std::uint32_t kPmuOverflowMask = 0xffU;
 
 constexpr std::uint8_t kArmv7MFullPcBytes = 4U;
-constexpr std::uint8_t kArmv7MAddressOffsetBytes = 2U;
 constexpr std::uint8_t kArmv8MMatchBytes = 1U;
 constexpr std::uint32_t kArmv8MMatchValue = 1U;
+
+/** @brief Returns whether a raw DWT address fragment width can be preserved. */
+static bool isSupportedAddressOffsetSize(std::uint8_t size)
+{
+  return size == 1U || size == 2U || size == 4U;
+}
 
 /** @brief Describes an invalid DWT event-counter payload. */
 static std::string invalidEventCounterMessage(const DwtPayloadPacket& payload)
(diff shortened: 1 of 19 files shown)


jreineckearm (commented):
codecov seems to complain about a missing coverage line.
I assume the rename from `offset` to `address` comes separately is what you were still planning to push to this PR? Or does it come separately.

jreineckearm (approved):
LGTM, tested the new ctrace executable for a bit and couldn't spot regressions.

Issue #2599, opened 2026-09-07
Copyright check fails inside a C++ string literal


brondani:
**Describe the bug**

The `global/copyright` check fails while scanning `tools/ctrace/test/unit/src/output/ctf/CtfMetadataWriterTests.cpp`, even though the file has a valid copyright and SPDX header.

The failure is caused by this C++ string literal:

```cpp
const auto returnEntry = xml.find("path=\"EXCEPTION_RETURN/*\" displayText=\"true\"");
```

The pinned `comment-parser==1.2.4` dependency interprets `/*` inside the string as the start of a block comment and raises an unterminated-comment parse error. This line was introduced by #2587. That PR's `global/copyright` check also failed, but the PR was merged, so later pull requests fail because the workflow runs the hook with `--all-files`.

**To Reproduce**

1. Check out a revision containing merge commit `40ac6215` from #2587.
2. Install the pre-commit dependencies used by the repository, including `comment-parser==1.2.4`.
3. Run:

   ```shell
   pre-commit run --all-files
   ```

4. Observe that `global/copyright` fails while `comment-parser` processes `CtfMetadataWriterTests.cpp`.

The same failure can be isolated by running the copyright checker against that file:

```shell
python scripts/check_copyright_notice.py tools/ctrace/test/unit/src/output/ctf/CtfMetadataWriterTests.cpp
```

**Expected behavior**

The copyright check should complete successfully for source files that contain valid copyright and SPDX headers.

**Environment**

- Workflow: `global/copyright`
- `pre-commit`: 4.5.1
- `comment-parser`: 1.2.4
- Reproduced on Linux with Python 3.10.12

**Additional context**

- Introducing PR: https://github.com/Open-CMSIS-Pack/devtools/pull/2587
- Introducing merge commit: `40ac6215d14bc2a45a61d58fe0e506de790298fe`
- Workaround commit in the PR history: `c217140e9c7a30fb05a3e83050b8dd7dd05d5fa4`
- The workaround was reverted by the final review commit: `0f8047d00c469d6deed66bc326ea75c7f715af49`

JonatanAntoni:
The package `comment-parser` could be updated to `1.2.5`.

soumeh01:
Fixed in https://github.com/Open-CMSIS-Pack/devtools/pull/2600

brondani:
[...]

PR #2600, opened 2026-09-08
Make copyright checks dependency-free

soumeh01:
##Fixes:
- https://github.com/Open-CMSIS-Pack/devtools/issues/2599

Parse leading C/C++ and shell comments directly, removing the magic and comment-parser CI dependencies.

## Checklist

### DIFF
--- a/scripts/check_copyright_notice.py
+++ b/scripts/check_copyright_notice.py
@@ -13,11 +13,64 @@ import argparse
 import os
 import sys
 import re
-import magic
-from comment_parser import comment_parser
 
 COPYRIGHT_TEXT = "Copyright (c) <ValidYear>"
 LICENSE_TEXT = "SPDX-License-Identifier: Apache-2.0"
+
+def extract_header(source: str, is_shell: bool) -> str:
+    """Return leading blank and comment lines, stopping at the first code line."""
+    header = []
+    in_block_comment = False
+    include_guard = None
+
+    for line in source.splitlines(keepends=True):
+        stripped = line.strip()
+
+        if in_block_comment:
+            header.append(line)
+            if "*/" in stripped:
+                in_block_comment = False
+                if stripped.split("*/", 1)[1].strip():
+                    break
+            continue
+
+        if not stripped:
+            header.append(line)
+            continue
+
+        if is_shell and stripped.startswith("#"):
+            header.append(line)
+            continue
+
+        if not is_shell and stripped.startswith("//"):
+            header.append(line)
+            continue
+
+        if not is_shell and include_guard is None:
+            guard_match = re.fullmatch(r"#ifndef\s+([A-Za-z_]\w*)", stripped)
+            if guard_match:
+                include_guard = guard_match.group(1)
+                header.append(line)
+                continue
+
+        if not is_shell and include_guard:
+            define_match = re.fullmatch(r"#define\s+([A-Za-z_]\w*)", stripped)
+            if define_match and define_match.group(1) == include_guard:
+                include_guard = ""
+                header.append(line)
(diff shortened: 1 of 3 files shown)


scripts/check_copyright_notice.py
```diff
@@ -13,12 +13,65 @@
...
+    """Return leading blank and comment lines, stopping at the first code line."""
+    header = []
+    in_block_comment = False
+    include_guard = None
+
+    for line in source.splitlines(keepends=True):
+        stripped = line.strip()
+
+        if in_block_comment:
+            header.append(line)
```
brondani: Exclude text after the block-comment terminator so trailing code cannot satisfy the notice checks.

```suggestion
            header.append(line if "*/" not in line else line[:line.index("*/") + 2])
```

scripts/check_copyright_notice.py
```diff
@@ -13,12 +13,65 @@
...
+
+        if not is_shell and include_guard:
+            define_match = re.fullmatch(r"#define\s+([A-Za-z_]\w*)", stripped)
+            if define_match and define_match.group(1) == include_guard:
+                include_guard = ""
+                header.append(line)
+                continue
+
+        if not is_shell and stripped.startswith("/*"):
+            header.append(line)
```
brondani: For single-line block comments, retain only the actual comment text before searching for notices.
```suggestion
            header.append(line if "*/" not in line else line[:line.index("*/") + 2])
```

brondani (commented):
> Model: **cmsis-dev-openai-proxy/gpt-5.6-sol**
> Metrics: 78s | 1,411 input tokens | 722 output tokens | 2,133 total tokens

## 1. Findings (ordered by severity: High, Medium, Low)

### Medium — Code after a block-comment terminator is searched as if it were a comment

**File:** `scripts/check_copyright_notice.py`  
**Lines:** 30 and 64

`extract_header()` appends the entire line before checking whether `*/` is followed by code. Although parsing then stops, `check_file()` still searches the already-appended trailing code.

For example, a line such as:

```c
/* header */ const char *notice = "SPDX-License-Identifier: Apache-2.0";
```

can satisfy the license check even though the SPDX identifier is not in a comment. The same issue occurs when a multiline block comment ends on a line containing trailing code. This weakens the correctness of a compliance check and differs from `comment-parser`, which returned comment text only.

**Concrete fix guidance:** When a block comment closes, append only the text through the first `*/`; do not include anything after the terminator in the extracted header. Apply this both while already inside a block comment and when processing a block comment that starts and ends on one line.

## 2. Suggested fixes per finding

### Block-comment trailing code

Truncate closing block-comment lines at the first `*/` before adding them to `header`. Keep the existing decision to stop parsing when non-whitespace follows the terminator.

The inline suggestions below apply this change at both append sites.

## 3. Final summary

The dependency removal is reasonable, but the new parser currently allows required notice text in trailing code to be treated as comment content. This should be fixed and covered by focused parser tests before merging.

## Inline suggestions

scripts/check_copyright_notice.py
soumeh01: [Addressed review comments](https://github.com/Open-CMSIS-Pack/devtools/pull/2600/commits/e6567d7bf19c7222cfc1f5c5251dd9fac75c9dee)

scripts/check_copyright_notice.py
soumeh01: [Addressed review comments](https://github.com/Open-CMSIS-Pack/devtools/pull/2600/commits/e6567d7bf19c7222cfc1f5c5251dd9fac75c9dee)

brondani (approved):

PR #2624, opened 2026-09-17
fix(ctrace): accept default TB captures and report skipped payload

thorstendb-ARM:
## Fixes

- TB/TB_* captures rejected when the optional `trace-format` declaration is missing or null.
- Premature failure on initial formatter payload without an assigned source ID.
- Missing visibility of skipped payload and a summary that described diagnostic records as decoded events.

## Changes

- Resolve the input format before route normalization: TB/TB_* defaults to formatted CoreSight, SWO to
  unformatted ITM. Explicit declarations retain precedence; input selection remains unambiguous.
- Tolerate initial unassigned formatter payload and account for payload skipped due to missing, NULL,
  reserved, or unconfigured source IDs, plus initial routed bytes before real hardware synchronization.
- Report skipped bytes consistently as CLI Info and timestamp-free CSV `info` rows, independently of
  `--type` and `--stream` filters. These annotations do not create synthetic routes or CTF events.
- Keep missing synchronization on an observed configured route as an end-of-input Error with non-zero exit,
  retaining completed diagnostics and healthy-route outputs. Genuine framing failures still abort outputs.
- Use packet-neutral skip wording and a summary of processed input bytes and trace/diagnostic records,
  rather than claiming that every record represents a decoded payload event.
- Extend regression coverage and update README, architecture MD/SVG, constraints, and output documentation.

## Validation

- Native macOS arm64 Release build passed.
- All seven ctrace CTest groups passed: 377 unit tests passed with five platform-specific skips,
  37 integration tests passed, plus fixture integrity and executable/manifest smoke checks.
- Regression coverage includes channel defaults and overrides, incomplete frames, exact byte accounting,
  chunk boundaries, late/missing synchronization, independent routes, filtering, output failures,
  legacy raw SWO golden outputs, and the reconstructed multi-source TB fixture.
- Three additional real TB captures were checked locally with `--csv`:

  | Raw size | Result |
  |---|---|
  | 64 bytes | No hardware SYNC: four skip Infos and one Error; exit 1. |
  | 4,160 bytes | Hardware SYNC only at the end: four skip Infos; no payload events; exit 0. |
  | 4,096 bytes | Hardware SYNC only at the end: three skip Infos; no payload events; exit 0. |

  The latter two captures have no routed payload after their first SYNC; they are not successful PC-sample decodes.
- Changed Markdown files, SVG XML syntax, and `git diff --check` passed.

## Scope and limitations

- No raw bytes are rewritten, no synthetic synchronization is injected, and no heuristic recovery CLI option is added.
- Counts describe deformatted payload bytes, excluding formatter-control bytes; this is not a general accounting
  mechanism for every possible decoder-recovery loss.
- Exactly one input remains required; formatted input still uses complete 16-byte memory-aligned frames.
- `trace-format` remains a provisional ctrace-private field, not a newly standardized producer requirement.
- No additional protocol decoders, dependency updates, or repository-global workflow changes.
- Cross-platform CI and coverage checks will run on this PR; local validation above is macOS arm64.

## Checklist

- [x] 🤖 This change is covered by unit and integration tests.
- [x] 🤹 Manual testing has been performed with real TB captures.
- [x] 🛡️ Security impacts considered: bounded accounting, unchanged input, and no invented synchronization.
- [x] 📖 Documentation updates are complete.
- [x] 🧠 Third-party dependencies and TPIP reviewed; no changes required.


tools/ctrace/docs/multi-source-design.md
```diff
@@ -0,0 +1,212 @@
...
+            TraceEvent flow
+           /              \
+   one combined CSV    one CTF bundle
+                      metadata + stream_<id> files
+```
+
+## Input and route identity
+
+Input format describes the effective capture bytes. A single source can be formatted, and a sink's name does not
+prove how its formatter was configured. The migration introduced the provisional `ctrace-run.trace-format` field
```
jreineckearm: `ctrace-run.trace-format` won't make it into the spec as per this yet to be discussed proposal: https://github.com/Open-CMSIS-Pack/cmsis-toolbox/pull/699
Main reason is to avoid polluting the trace source configuration in `ctrace-run` with trace communication configuration (formatter is part of the trace sink).

However, the current default behavior is what we need for the first release (formatter on for `TB` and `TB_<name>` files, off for `SWO` files)

tools/ctrace/docs/THIRD_PARTY_NOTICES.md
```diff
@@ -5,6 +5,11 @@ release archive contains the unmodified license text for each component under
 `THIRD_PARTY_LICENSES`. OpenCSD's retained source-file copyright notices are
 provided in `THIRD_PARTY_LICENSES/OpenCSD-NOTICE.txt`.
 
+`scripts/package-release.sh` packages the five platform binaries, the devtools
```
jreineckearm: Does this note belong here? Kind of interrupts the license notes with unrelated checksum stuff.

jreineckearm (commented):
Two comments from code review. Rest of the code looks good.
Taking for a manual test now.

jreineckearm (dismissed):
Testing looks good, I'd be fine with keeping the `trace-formatter` provision in for the time being. But as mentioned in my comment I don't see it becoming part of the spec.

tools/ctrace/docs/multi-source-design.md
thorstendb-ARM: Clarified in 55474d85: `trace-format` remains a temporary private override, with no assumed standardization. The TB/named-TB and SWO defaults and implementation are unchanged.

tools/ctrace/docs/THIRD_PARTY_NOTICES.md
thorstendb-ARM: Moved the packaging/checksum paragraph to `architecture.md` (Build and CI structure) in 55474d85. The third-party notices retain the licensing information.

jreineckearm (approved):
Thanks for the updates, reviewed them. No need to retest as pretty much all was in documentation files. Good to merge.

PR #2629, opened 2026-09-21
fix(ctrace): support PC sampling status markers

thorstendb-ARM:
## Fixes

- Treat the valid Armv8-M one-byte DWT PC-sampling payload `0xFF` as `Trace prohibited`, not a decoder error.
- Add the missing `CPU Sleeping` CSV note for the one-byte `0x00` marker, following the
  [PC-sampling marker specification](https://open-cmsis-pack.github.io/cmsis-toolbox/Experimental-Features/#pc-sampling-markers).
- Avoid empty Trace Compass analyses: the consumer rejects an XML `stateProvider` without graphical content.

## Changes

- Replace the semantic sleep boolean with explicit PC, sleep, and trace-prohibited states. Four-byte payloads remain
  PC addresses, including `0x000000FF`; unsupported payloads remain errors. Arbitrary raw `0xFF` bytes are not markers.
- Preserve route, timestamp, quality, and the existing `pcsample` selection. CSV leaves the PC column empty for markers
  and writes the specified note.
- Add CTF event `PC_SAMPLE_PROHIBITED` (ID 10) with the common sample-quality fields. The existing `PC_SAMPLE` binary
  layout, event IDs, and profile version remain unchanged; its state field still determines a zero-or-one PC array.
- Close any open route-local Sleep interval at the prohibited marker without inferring running state or a prohibited
  duration. Omit companion XML when no graphical topics remain, remove stale XML, and retain the CTF event table.
- Add a synthetic marker fixture, negative/routing/filter tests, SWO and formatted TB integration coverage, a real
  Babeltrace consumer check, and documentation of the schema and Trace Compass acceptance.

Scope is limited to ctrace. No OpenCSD dependency or shared workflow changes.

## Validation

- macOS Release: all 7 CTest entries passed.
- Linux/GCC 13: all 8 CTest entries passed, including 358 unit tests, 34 integration tests, and Babeltrace 2.0.5.
- CI-compatible source-line coverage: **6410/6410 lines (100%)**.
- Actual Trace Compass Server 0.17.0 / TSP 0.6.0: PC → Sleep → Prohibited → PC at 1/3/6/10 microseconds; Sleep spans
  exactly `[3, 6)` microseconds. Marker-only input retains its table event, emits no empty XML or graphical view, and
  removes stale XML. Testing used an isolated server instance.
- Fixture integrity, Markdownlint, and `git diff --check` passed.

## Checklist

- [x] 🤖 This change is covered by unit tests.
- [x] 🤹 Manual consumer testing has been performed.
- [x] 🛡️ Security impacts have been considered; marker handling neither reconstructs a hidden PC nor bypasses payload validation.
- [x] 📖 Documentation updates are complete.
- [x] 🧠 Third-party dependencies and TPIP reviewed; no updates required.


jreineckearm:
@JonatanAntoni , could you please run a copilot review?

tools/ctrace/src/decode/DwtPacketDecoder.cpp
```diff
@@ -180,21 +197,21 @@ std::vector<TraceEvent> DwtPacketDecoder::decodeExceptionTrace(const DwtPayloadP
...
             ", value " + std::to_string(payload.value) +
-            "; expected a 4-byte PC or a 1-byte zero sleep indication",
+            "; expected a 4-byte PC or a 1-byte marker (0x00: CPU Sleeping, 0xff: Trace prohibited)",
         std::nullopt,
         std::nullopt,
     }));
     return output;
   }
-  output.push_back(makeDwtEvent(payload, PcSampleTraceEvent{payload.value, isSleeping}));
+  const auto pc = kind.value() == PcSampleKind::Pc ? payload.value : 0U;
```
jreineckearm, line 213: Is value `0` the only indicator that this is a marker a not a valid PC? It can be a valid code address ever since VTOR became programmable.

tools/ctrace/test/data/trace-pc-sample/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+| Bytes | Meaning | Timestamp after local timestamp |
+| --- | --- | --- |
+| `00 00 00 00 00 80` | ITM hardware synchronization | 0 |
+| `17 34 12 00 08 10` | PC `0x08001234`, local increment 1 | 1 |
+| `15 00 20` | CPU sleeping, local increment 2 | 3 |
+| `15 FF 30` | Trace prohibited, local increment 3 | 6 |
+| `17 78 56 00 08 40` | PC `0x08005678`, local increment 4 | 10 |
+
+The YAML declares a 1 MHz clock and prescaler 1. Integration tests use the
```
jreineckearm: Nit: Describing the raw content makes sense. Is the rest special for this test? Or does it also apply to other test data without README? Just wondering how far to drive this level of detailed documentation.

tools/ctrace/test/integration/src/ValidateFixtureIntegrity.cmake
jreineckearm, line 1: Nit: Are these files manipulated during tests? Just wondering why we need to maintain hashes, especially for the non-binary files.

tools/ctrace/README.md
```diff
@@ -24,6 +24,12 @@ If the directory follows them, terminate option parsing explicitly, for example
 `ctrace --type itm dwt -- .trace`. With no output option, `ctrace` validates and decodes the capture without writing
 output files. Run `ctrace --help` for the current option details.
 
+The `pcsample` selection includes PC values and the
```
jreineckearm: Probably worth to remove this. As I can see, this is the only time we describe a particular trace packet. If I missed others (maybe overflow or error rows?), then could make sense to move there.

jreineckearm (dismissed):
LGTM, cannot manually test though as I don't have reproducers for the special PC sample packets.

Please see some comments, can be addressed in following PRs if deemed worthy.

tools/ctrace/src/decode/DwtPacketDecoder.cpp
thorstendb-ARM: Payload size distinguishes them: four-byte `0x00000000` is a valid PC; only one-byte `0x00` means Sleep. Added decoder/CSV and raw-byte pipeline regression coverage in 7cd2d1b0; no decoder change needed.

tools/ctrace/README.md
thorstendb-ARM: Removed the packet-specific paragraph from the main README in 7cd2d1b0. The details remain in `constraints.md` and `ctf-format.md`.

tools/ctrace/test/data/trace-pc-sample/README.md
thorstendb-ARM: Trimmed in 7cd2d1b0: kept fixture provenance, byte sequence and clock/prescaler; replaced duplicated test details with links.

tools/ctrace/test/integration/src/ValidateFixtureIntegrity.cmake
thorstendb-ARM: Tests use working copies; hashes guard checked-in inputs and reference outputs against unintended changes. Removed Markdown hashes and the redundant fixed file count in 7cd2d1b0; test-data/configuration checks remain.

jreineckearm: Thanks for clarifying and updating the docs.

tools/ctrace/src/decode/DwtPacketDecoder.cpp
jreineckearm: Thanks for clarifying, didn't consider the size part when looking at it.

jreineckearm (approved):
Thanks for the updates, good to merge.

